PR #16, opened 2016-10-14
a first unit-test proposal

alanw0:
Stefan,

This is a commit in my fork of Nalu which contains an initial unit-test proposal.

I'll stop by and see you in person.

Alan


spdomin:
As discussed, let's move forward with a more foundational first unit test, e.g., creating an element and checking it's coordinates. 

Then, we can integrate the algorithm design with unit testing to make the overhead of calling a Nalu algorithm consistent with unit testing design concepts.

PR #17, opened 2016-10-14
Basic unit-test infrastructure.

alanw0:
Demonstration unit-test creates a mesh with 1 element per processor,
verifies that each element is a unit cube by looking at nodal coordinates.


spdomin:
Well done...

PR #18, opened 2016-10-17
Abl bcs

mbarone81:
This is the base ABL surface boundary condition implementation.  Comments are solicited, esp, from @spdomin  and @mchurchf .  Current nightly test suite passes on Linux blade with this branch.  Neutral, unstable, and stable ABL simulations run without issue on a 3M element test mesh.


spdomin:
@mbarone81, did you resolve the intel-based issues? I will have this done by COB - possibly later in the day.

mbarone81:
@spdomin, yes, these were being caused by a division by zero issue within the compute_utau routine.

spdomin:
Okay, overall this looks nice. I only found three possible changes...

First, when the user species heat_flux, the Enthalpy equation system already registers a field names heat_flux_bc. We should use the same one for the momentum ABL and drop the heat_flux_bc_mom.

Second, I think that there is a problem with your management of gravity. Well, technically, you just added a new option, Gravity. However, we already have one in the code base, e.g., std::vector<double> soGravity = realm_.solutionOptions_->gravity_; As you can see, this is a vector quantity. If you use this option, which I think you should, then you may need to make some code changes to support a vector gravity and not a single scalar.

Third, place the ABL profile in user_functions.

All these are minor... Looks great. Since you followed the previous implementation, this formulation will work for low-order element- and edge-based in addition to higher order element-based.

mbarone81:
@spdomin, I have addressed the first two of your requested changes.  As we discussed, the ABL profile class that I created is a helper class that provides the Monin-Obukhov functions necessary for the ABL wall function, and should be retained as 'normal' source files.  I re-ran nightly tests on Linux blade and re-ran the 3M element ABL case for positive, negative, and zero heat flux.

spdomin:
I fixed up the gravity comment and pushed manually. Get the rest going and we will be set.

mchurchf:
I went ahead and pulled this and recompiled, but I get an error:

libnalu.a(LowMachEquationSystem.C.o): In function `sierra::nalu::MomentumEquationSystem::register_wall_bc(stk::mesh::Part*, stk::topology const&, sierra::nalu::WallBoundaryConditionData const&)':
LowMachEquationSystem.C:(.text+0xd365): undefined reference to `sierra::nalu::AssembleMomentumABLWallFunctionSolverAlgorithm::AssembleMomentumABLWallFunctionSolverAlgorithm(sierra::nalu::Realm&, stk::mesh::Part*, sierra::nalu::EquationSystem*, bool const&, double const&, double const&, double const&)'
LowMachEquationSystem.C:(.text+0xd3c1): undefined reference to `sierra::nalu::ComputeABLWallFrictionVelocityAlgorithm::ComputeABLWallFrictionVelocityAlgorithm(sierra::nalu::Realm&, stk::mesh::Part*, bool const&, double const&, double const&, double const&)'


By the way, is the best way for me to email issues like this?

Matt

Matthew J. Churchfield
Senior Engineer | National Wind Technology Center

National Renewable Energy Laboratory (NREL)
15013 Denver West Parkway | Golden, CO 80401
303-384-7080 | F: 303-384-6901
matt.churchfield@nrel.gov<mailto:matt.churchfield@nrel.gov> | www.nrel.gov<http://www.nrel.gov/>

mbarone81:
@mchurchf and @spdomin let me pull a clean copy and see if I can replicate...

spdomin:
@mchurchf this is probably due to the fact that you did not reconfigure Nalu at the build step. As such, your Cmake files are out of date. Go to /build, clean and re-config Nalu and build again.

mbarone81:
@mchurchf I believe this must be the case...I just pulled from master and rebuilt successfully

mchurchf:
@mbarone81 and @spdomin, yes, that was the problem.  All is fine now.

PR #19, opened 2016-10-17
Prototype algorithm design to remove resize

spdomin:
The objective of this project is to remove the usage of resize in the
algorithm::execute() in favor of a homogeneous collection of parts on which
the algorithm operates.
- Modify Algorithm class to manage a map<string, Algorithm*>
- LowMachEOS now creates a set of CourantReyAlgorithms for each
  topology,
  
  // non-solver CFL alg
  const std::string cflAlgName = "cflAlgName_" + baseSCSName;
  std::map<std::string, Algorithm *>::iterator its
    = cflReyAlgDriver_->algorithmMap_.find(cflAlgName);
  if ( its == cflReyAlgDriver_->algorithmMap_.end() ) {
    AssembleCourantReynoldsElemAlgorithm *theAlg
      = new AssembleCourantReynoldsElemAlgorithm(realm_, part, meSCS);
    cflReyAlgDriver_->algorithmMap_[cflAlgName] = theAlg;
  }
  else {
    its->second->partVec_.push_back(part);
  }

The above results in multiple algorithms (each with many potential parts)
of the same topology. The core design would be that one algorithm type
holds all of the parts which can be heterogeneous.

Notes:

a) kind of extensive - I think...
b) does this move towards the unit test approach to remove a realm?
c) Tricky... Algorithms, since they once handled all of the parts
   now operate on a possible subset of parts. As such, calls such as
   the following are tricky:

  // parallel max
  double g_maxCR[2]  = {};
  stk::ParallelMachine comm = NaluEnv::self().parallel_comm();
  stk::all_reduce_max(comm, maxCR, g_maxCR, 2);

  // sent to realm
  realm_.maxCourant_ = g_maxCR[0];
  realm_.maxReynolds_ = g_maxCR[1];

  You see, we now need something like the following:

  // before the execute
  realm_.maxCourant_ = 0.0;
  realm_.maxReynolds_ = 0.0;
  momentumEqSys_->cflReyAlgDriver_->execute();

  // sent to realm; at the end of the Alg class
  realm_.maxCourant_ = std::max(realm_.maxCourant_, g_maxCR[0]);
  realm_.maxReynolds_ = std::max(realm_.maxReynolds_, g_maxCR[1]);

  So... Conclusion is that Algorithms can no longer be extected to
  run without an AlgorithmManager to organize the pre and post-work.

### DIFF
--- a/include/Algorithm.h
+++ b/include/Algorithm.h
@@ -45,6 +45,7 @@ public:
   virtual void pre_work() {}
 
   Realm &realm_;
+
   stk::mesh::PartVector partVec_;
   std::vector<SupplementalAlgorithm *> supplementalAlg_;
 };
--- a/include/AlgorithmDriver.h
+++ b/include/AlgorithmDriver.h
@@ -33,6 +33,7 @@ public:
 
   Realm &realm_;
   std::map<AlgorithmType, Algorithm *> algMap_;
+  std::map<std::string, Algorithm *> algorithmMap_;
 };
 
 } // namespace nalu
--- a/include/AssembleCourantReynoldsElemAlgorithm.h
+++ b/include/AssembleCourantReynoldsElemAlgorithm.h
@@ -23,11 +23,21 @@ public:
 
   AssembleCourantReynoldsElemAlgorithm(
     Realm &realm,
-    stk::mesh::Part *part);
+    stk::mesh::Part *part, 
+    MasterElement *meSCS);
   virtual ~AssembleCourantReynoldsElemAlgorithm() {}
 
   virtual void execute();
-  
+
+  MasterElement *meSCS_;
+  const int nodesPerElement_;
+  const int numScsIp_;
+  const int *lrscv_;
+
+  const stk::mesh::BulkData *bulkData_;
+  const stk::mesh::MetaData *metaData_;
+  const int nDim_;
+
   const bool meshMotion_;
 
   VectorFieldType *velocityRTM_;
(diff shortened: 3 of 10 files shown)


spdomin:
Need some feedback on this prototype:

@alanw0, @rcknaus, @crtrott, @marchdf, @jclausen and @NaluCFD/ngp

alanw0:
Perhaps we could keep the "top level" algorithm, e.g., AssembleCourantReynoldsElemAlgorithm, but now this top-level algorithm holds a collection of algorithms templated on topology, e.g., AssembleCourantReynoldsElemAlgorithm<topology>. Then the top-level algorithm handles executing each of the templated algorithms, and assembling the g_maxCR quantity before giving it to realm.

The top-level algorithm could hold the same selector as before, and each templated algorithm can iterate over entities specified by (selector & topology-part), which produces non-overlapping subsets of elements.

spdomin:
@alanw0, yes. Would this top level algorithm look like the AlgorithmDriver class? Some ideas that came back also include helper (templated) methods to look up the map.

alanw0:
I guess it would look like AlgorithmDriver, executing the templated algorithms and assembling their result quantities.

crtrott:
I have to think about this a bit harder to understand whether or not there are performance issues with this approach lurking in the dark.

spdomin:
@crtrott, yes. Please do. My only concern was that for a typical hybrid mesh (hex/tet/pyr/wedge) we would have four instantiations of the algorithms as apposed to a single one with one resize at the beginning of the execute. Towards the later, we could easily have the realm keep track on the max size of master element that comes through.

However, this was a low-cost prototype to start the dialogue on alternative designs. Perhaps in the end, the look-and-feel of your Nalu Kokkos implementation will be good. I just want to test a few other concepts.

PR #20, opened 2016-10-17
Optimize the handling of ghost elements in NonConformalManager.

alanw0:
Added a function to NonConformalManager called
computePreciseGhostingLists. This function starts from the elemsToGhost_
vector, figures out which of those are already ghosted, and figures out
which receive-ghosts no longer need to be ghosted. Then the stk
change_ghosting function can be called with a minimal list of elements
to ghost, and a list of ghosts that can be removed from ghosting.

This change removes the need to destroy and re-ghost everything, and
also reduces the number of calls to stk modification_end.
For the test dgNonConformalEdgeCylinder this reduces runtime from
~71s to ~53s on linux blade.

I will also propose a patch to the stk team which further reduces
runtime for this test to ~49s. (The stk patch improves efficiency
within the stk ghosting operation.)


spdomin:
@alanw0 as discussed, let's change some syntax and touch base tomorrow.

The current coordinates field forms the off-processor bounding box. This will be accurate. However, I can not see how the fine search will be correct if the previously ghosted element nodal coordinates are not communicated in your new routine. I think that this is a mystery we need to figure out.

In the meantime, adding a parallel_communicate to the exposed_area_vector in all of the NonConfAlg algorithms in addition to a parallel_communicate on realm_.current_coordinates() after the manage_ghosting() call seems to be 100% correct.

alanw0:
@spdomin, adding exposedAreaVec_ to ghostFieldsVec_ in the NonConfAlgs does solve the diff. I will soon push a modification to the pull request.

spdomin:
@alanw0 you also need to add a parallel_communication after the coarse search and ghosting of the field name "realm_current_coordinates()" and just prior to the fine search.

alanw0:
@spdomin, ok, those changes are in.

spdomin:
@alanw0 good news thus far. My longer running cylinder case was around 15% faster with your new addition (no diffs).

At time step 100:

Old:  NonConformal alg will ghost a number of entities: 81630
New: NonConformal alg will ghost a number of entities: 740 and remove 7637 entities from ghosting.

I have a large case ready to run now... After that is clean, I will accept the pull request.

alanw0:
@spdomin sounds pretty good. I guess the improvement in run-time is problem specific. I can keep working on that for more improvements if it seems valuable. Or, perhaps make this ghosting scheme available to other algorithms in nalu.

spdomin:
Okay, it looks like the larger run is less convincing on timings. I will have more data in a day or so. Good news: there are no diffs on these larger runs.

I see the diagnostic now... Basically, it is a change status... How many "new" are in elemsToGhost_ and the change in recvGhostsToRemove_. However, do we have a precise total of current ghosted entities? That might be a good thing to see. 

Any chance that at scale this new implementation can be slower?

spdomin:
@alanw0 do you think that when the size of the elements to ghost is really really large, the STL sorts, etc. are dominating? Also, the interface to coarse_search is a std::vector. However, I wonder if it would be better to use another STL (map, set, etc.) to make sure the list is unique and then copy into a vector? That or change the STK interface:)

alanw0:
@spdomin good question, I think I can explore that with an artificial mesh, and see how the cost of the different pieces changes with number of entities.
Also I notice that the nightly test dgNonConformalFluids doesn't seem to benefit from the new ghosting approach, so I may profile that and see what's happening.

spdomin:
@alanw0 results for the 256 6 mil Sky run are in...
New: STKPERF: Total Time: 48153.9
Old: STKPERF: Total Time: 48024.8

Close, but the new is a bit slower.

Also, unfortunately, I am working in these NonConfAlgorithms to deal with the opposing unit normal. As such, I actually removed the HACK usage of the exposed_area_vec for the opposing faces. Therefore, those new parallel communicate calls including exposedAreaVec_ are no longer required. The new lags that I will push may conflict with yours. Sorry... This has been a long standing issue for the sliding mesh scheme:) Just revert your local changes.

alanw0:
Ok, good data. I have a unit test where I can arbitrarily increase mesh size and size of ghost lists. This morning I will figure out what's going on. No problem about the conflict.

alanw0:
@spdomin I have a new candidate for the ghosting optimizations. I will run the cube6M problem on skybridge to make sure it's good before I ask you to reconsider a merge. Which compiler/version did you build with for skybridge?

spdomin:
@alanw0 I use the intel 14 suite. I think that I sent you the full set of modules.

One thing that I realized as well (although I do not think it is the main issue here) is that my local runs used the latest Trilinos while my Sky runs used an older version..

alanw0:
I had to delete and re-create my fork. I'm going to close this pull request, and open a new one when I get my ghosting changes re-verified.

PR #26, opened 2016-10-26
Add unit-tests that show simple kokkos usage.

alanw0:
These unit-tests don't interact with or modify nalu algorithms. They
show some simple kokkos operations such as creating views,
performing parallel_for loops, etc.

### DIFF
--- a/unit_tests.C
+++ b/unit_tests.C
@@ -7,6 +7,11 @@
 
 #include <gtest/gtest.h>                // for InitGoogleTest, etc
 #include <mpi.h>                        // for MPI_Comm_rank, MPI_Finalize, etc
+#include <Kokkos_Core.hpp>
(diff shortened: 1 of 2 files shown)


spdomin:
@alanw0 did you have to modify any config files to pull in this dependency or do we have it by virtue of hooking up Tpetra?

alanw0:
@spdomin no config changes were necessary, we already had the kokkos libraries. You're probably correct that they came in with Tpetra.

If you want to wait a little bit on this pull, I have a follow-up commit nearly ready to push which adds a deep_copy and clarifying comment to the unit-tests, which are important for semantic correctness even though the unit-tests as they were, are correct for OpenMP threading (but they wouldn't have been correct for Cuda).

alanw0:
@spdomin I have now pushed the second commit, these basic-kokkos unit-tests are ready to go.

PR #27, opened 2016-10-28
Start adding some master element unit tests

rcknaus:
Starts adding in some unit tests for the master element functions in Nalu.  Tests interpolation and gradient operations on reference hex elements.

@alanw0 , @spdomin


spdomin:
okay, @alanw0 feel free to generalize.

PR #28, opened 2016-10-31
Demonstrate unit-test fixture.

alanw0:
Made a class 'MasterElementHexSerial', which declares/creates the mesh,
then changed 'TEST' functions to 'TEST_F', which are actually member
methods of the MasterElementHexSerial class so they can refer to the
mesh and other members of that class.

In this case it saves a few lines of duplicated code for each test,
not huge, but demonstrates a handy gtest feature which can remove
a lot of duplication in some cases.

@rcknaus and @spdomin make sure you're happy with the readability of this. If not, no big deal.
If looking at the diff that github displays, click the 'split' view. This diff is hard to follow in 'unified' mode.

### DIFF
--- a/unit_tests/UnitTestHexMasterElements.C
+++ b/unit_tests/UnitTestHexMasterElements.C
@@ -224,118 +224,90 @@ void check_derivatives(
   }
 }
 
+class MasterElementHexSerial : public ::testing::Test
+{
+protected:
+    MasterElementHexSerial()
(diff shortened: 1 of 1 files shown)


rcknaus:
Looks good to me.  Thanks.

spdomin:
This looks good to me. I will let you and RK iterate on the design specifics...

PR #30, opened 2016-11-01
Unit-test comparing a calculation with and without kokkos views.

alanw0:
Stefan, take a look at the 2 unit-tests in this commit, if you don't mind. I've tried to mimic the calculation you proposed, however I don't know how to assert correctness.
The two tests can be run like this:
mpirun -np 1 unittestX --gtest_filter=Hex8Mesh.indexing_raw_arrays
mpirun -np 1 unittestX --gtest_filter=Hex8Mesh.indexing_views

gtest reports runtime. I haven't run these on any interesting platforms, but for a release build on the linux blade it appears as though the views version is a hair faster than the raw-arrays version.

### DIFF
--- a/unit_tests/UnitTestKokkosViews.C
+++ b/unit_tests/UnitTestKokkosViews.C
@@ -37,7 +37,7 @@ protected:
 
     ~Hex8Mesh() {}
 
-    void fill_mesh(const std::string& meshSpec = "generated:30x30x30")
+    void fill_mesh(const std::string& meshSpec = "generated:100x100x100")
     {
       unit_test_utils::fill_hex8_mesh(meshSpec, bulk);
     }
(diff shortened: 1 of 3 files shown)


alanw0:
Oops, I accidentally included a CMakeLists.txt change which I can remove. I had turned off the -Werror flag to do debug builds, I didn't mean to push that.

alanw0:
ok, I amended the commit to remove the CMakeLists.txt change.

alanw0:
One last comment: zeroing the p_dpdxIp value at line 101 is wrong (and similarly in the second test). I will remove that.

alanw0:
Added a commit that allows easy switching of the bucket-loop scheme. Encapsulates the kokkos boiler-plate code that is necessary for thread-team nested parallelism.

spdomin:
@rcknaus can you look at this and provide an opinion? I am swamped...

rcknaus:
The code looks good as a test to keep record of the cost of switching to Kokkos views for Nalu.   A couple of points:

1) I think you want to just index dpdxIp by ip in the innermost loop.  This would be the gradient flux of pressure across the subcontrol face associated with the integration point "ip" (one integration point per subcontrol surface for P=1).  Testing correctness for that would be fairly difficult.  If we do the full flux differencing (using the "lrscv -- left/right subcontrol volume" mapping), then putting in a linear pressure and expecting "0"  out would work.  I can make the changes required and push them back to you if you want.

2) Do we want to try to split these types of tests (where we care primarily about timing) to a micro-benchmark suite?

spdomin:
Okay, I accepted the pull request. I can iterate with @rcknaus

alanw0:
@rknaus thanks for the comments. I wasn't sure on the math, I believe you're right about the dpdxIp field. Your idea for testing correctness sounds good to me. If you are willing to make that change, that would be great. 

Regarding splitting timing tests to a benchmark suite, that sounds good, but I'm not sure specifically what that means. Do you mean a separate executable? A separate repository? Stay in the unit-test executable but they only run if a particular command-line option is provided?

spdomin:
I thin RK resolved this with his latest push. Thanks A for the initial condition.

PR #33, opened 2016-11-04
Modify KokkosView unit test to compute the discrete Laplacian

rcknaus:
I changed the unit test to compute the discrete Laplacian, and compared the result with the exact result for a quadratic polynomial.  The max error for this test on my local machine is about 1e-11, and I put the tolerance at 1e-10.

### DIFF
--- a/unit_tests/UnitTestKokkosViews.C
+++ b/unit_tests/UnitTestKokkosViews.C
@@ -7,6 +7,7 @@
 #include <stk_mesh/base/BulkData.hpp>
 #include <stk_mesh/base/CoordinateSystems.hpp>
 #include <stk_mesh/base/Field.hpp>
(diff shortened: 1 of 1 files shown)


spdomin:
@rcknaus, the np2 unit test now seems to be failing.

/scratch/spdomin/nightlyBuildAndTest/Nalu/unit_tests/UnitTestKokkosViews.C:229: Failure
The difference between *stk::mesh::field_data(*discreteLaplacianOfPressure, node) and exactLaplacian is 495.4
5533695517901, which exceeds tol, where
*stk::mesh::field_data(*discreteLaplacianOfPressure, node) evaluates to 494.61135609330972,
exactLaplacian evaluates to -0.84398086186930299, and
tol evaluates to 1e-10.

PR #34, opened 2016-11-07
Fix KokkosView unit test in parallel

rcknaus:
Fixes the selectors to get the test working in parallel.


alanw0:
Looks good to me!

PR #37, opened 2016-11-17
Enable ctest

marchdf:
Please let me know what you think of this CTest setup. It runs nightly on peregrine, and checks both an gcc and Intel build. Results are posted to [this CDash server](http://my.cdash.org/index.php?project=Nalu).  See also the README in the cmake folder for more details on how it runs. 

The only thing I would like to remove at some point is a hardcoded link to a do-configNalu script I have for peregrine (gcc and intel). Maybe with the ongoing work on SPACK this might make this possible.


spdomin:
I like the changes to support Ctest. Do we really need all of these specialized config files inside the repository though? I was always thinking that we would supply some bare bones config files that people customize without the need to push back to the repos.

I guess the same thing with the ctest scripts. Should we have a specific scrip for Peregrine within the repos or just manage this outside the repos in some buildAndTest directory?

marchdf:
I actually got rid of the specialized do-config* scripts (they were committed by mistake early on) and they shouldn't appear in the final commit.  I am not sure how to get rid of them in the git history but I can try to do that if you want. 

The ctest scripts themselves should be independent of Peregrine and applicable to any system/build (except that it does call a do-configNalu script that is specific to the system, eventually I want to replace that with a SPACK install call).

spdomin:
Okay, good. I generally squash my commits when I push to that the history looks cleaner. Alternatively, I can be really rigorous and make sure that each commit corresponds to a "clean" part of the code.

Squashing makes bisecting easier as well since sometimes the code at a given commit may not even build..

>git rebase --interactive HEAD~4

marchdf:
Ok I will work on cleaning the history. Closing this pull request for now.

PR #39, opened 2016-11-18
Enabled CTest/CDash testing suite.

marchdf:
Cleaned up the commit history compared to the previous (canceled) pull request. Let me know what you think.


spdomin:
@marchdf can you send a message to the team instructing them how to update the ctest files when a new Rtest is added?

PR #45, opened 2016-11-30
Update ctest prepare after Epetra deprecation and new xml files 

marchdf:
xml files were added for 16 of the tests during the Epetra deprecation. CTest is now aware of these new files.


spdomin:
Sorry about forgetting about this... I guess we will need to make a decision as to how we run our tests soon. This way, we do not need to support both the Nalu script and the Ctest path...

PR #52, opened 2017-01-03
Optimizations for custom ghosting.

alanw0:
Do an incremental update of entities to be ghosted, rather than
destroying and re-creating the ghosting. Also avoid a call
to mod-begin/mod-end in some cases.

These changes are shown to improve overall run-time by as much as
25% on a nightly test (dgNonConformalEdgeCylinder) and by about
7% on the cube6M problem on 256 procs of skybridge (total runtime
goes from ~46000s to ~42800s).

### DIFF
--- a/include/NonConformalManager.h
+++ b/include/NonConformalManager.h
@@ -52,7 +52,6 @@ class NonConformalManager {
   ~NonConformalManager();
 
   void initialize();
-  void manage_ghosting();
 
   Realm &realm_;
   const bool ncAlgDetailedOutput_;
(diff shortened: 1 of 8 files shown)


spdomin:
Sorry, is this up-to-date? In other words, did you pull? I see some algorithm changes that I thought I made months ago, however, are showing as diffs...

alanw0:
I did pull. If you are referring to the few places where I added edgeAreaVec_ to the ghost-fields vec, I believe those are necessary for correctness. I can re-run the nightly tests without those changes to confirm if you like, but last time I recall running the nightly tests without those, there were some diffs.

spdomin:
Nice changes... I pushed this and will close with thanks to @alanw0

PR #63, opened 2017-01-31
Replace old stk::mesh::skin_mesh with newer stk capability.

alanw0:
The newer stk::mesh::create_exposed_block_boundary_sides is more
robust than skin_mesh, correctly handles a mesh which caused
skin_mesh to issue errors/warnings.

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -94,7 +94,7 @@
 #include <stk_mesh/base/MetaData.hpp>
 #include <stk_mesh/base/Comm.hpp>
 #include <stk_mesh/base/CreateEdges.hpp>
-#include <stk_mesh/base/SkinMesh.hpp>
+#include <stk_mesh/base/SkinBoundary.hpp>
 
 // stk_io
 #include <stk_io/StkMeshIoBroker.hpp>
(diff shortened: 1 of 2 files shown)


alanw0:
Stefan, sorry for the delay on this... This pull request contains the change to use the newer more robust stk skinning capability. As I noted previously, it correctly handled the mesh from Ganesh which the old skinning capability was not handling.

spdomin:
@alanw0, did you work through the overset use case of skinning? Does performance also increase in the new method?

alanw0:
@spdomin I did get it working for the overset cases that are in the nightly test suite. On the stk team we previously measured performance of the new skinning capability to be better than the old, but I didn't re-check the performance on nalu cases now. Would you like me to measure and report back?

spdomin:
I will reach out to the @NaluCFD/overset team... Has anyone benchmarked the overset overhead as of yet? I would like to quantify the cost of skin_mesh and parallel search costs. The performance test, oversetHybrid may be a good test when mesh motion is activated, however, omega is set to zero. @alanw0, let's focus on Kokkos/Nalu algorithm work:)

skennon10:
?I have not benchmarked overset overhead yet, but, if all goes well with the FEM-patch approach, I hope to be able to remove some of these costs and thus will need to benchmark to compare.


________________________________

PR #67, opened 2017-02-07
Introducing a documentation framework using Sphinx, Doxygen, and Doxylink

jrood-nrel:
This pull request would add in a documentation framework which I have building both locally and on readthedocs.org. The example documentation can be seen at http://nalu.readthedocs.io which is automatically generated after each push to the repo. The documentation content itself has been cobbled together (though thanks to Ganesh I have the theory manual included), but this is about the framework at the moment. I feel the minimum documentation would be for developers to annotate the code with Doxygen compatible comments going forward, to aid in automatically-generated developer-type documentation. In the future, a more user-facing and human-created documentation can be implemented using the Sphinx capabilities. Maybe we can discuss this at the next stand-up, and if people are on-board, we can merge it and connect readthedocs to this main repo instead of my personal one.


jrood-nrel:
Ok great! Thanks! Yes I can get the verification manual in as well (the theory manual I have in is out of date at the moment you will notice). I'm not sure how hard that would be, but Ganesh generated the theory manual from the LaTeX code, so I can just find out how he did it. I also hate to ask people to change their workflow, but going forward I am hoping that NaluDoc will essentially be replaced by the Sphinx documentation which can be seen in the new `docs/index.rst` and `docs/source` files/directories. Sphinx and Readthedocs are supposed to be able to generate PDFs automatically, but I haven't been able to get it to work yet, but it should be possible. To be clear, this documentation framework is completely separate from NaluDocs at the moment.

The update process basically involves annotating the source code with Doxygen-type comments and committing and pushing, then it hooks to Readthedocs and is updated automatically (Readthedocs checks out the code and builds the documentation itself). Or a change is made to the Sphinx documentation in `docs/source` and committed and pushed, then Readthedocs updates itself automatically in the same manner. I can try to set some examples of commenting for Doxygen now that the framework is in. You can use markdown and latex-type math directly in the code now for developer documentation.

For the class descriptions, it's very likely I have the Doxygen configuration screwed up somehow, and I will continue to iterate on issues like this in the documentation itself going forward.

My next step will be to transfer the Readthedocs link from my forked Nalu repo to this one so that when pushes occur here, Readthedocs will build from here, and I'm not sure how to do that yet. So until then, the nalu.readthedocs.io will be stale.

spdomin:
I think I misunderstood this pull request, however, not to worry, we can let it stand:)

Where is the theory manual to be developed now? I think that we guaranteed a duplication of files now, right? That is, unless we deprecated NaluDoc? Is that what you are planning? We push theory and verification (even user manual) to NaluCFD/Nalu?

jrood-nrel:
I apologize if I was hasty in this. I didn't mean to sneak it through or anything :-).

So my idea was to win you over by giving an example of what is possible to do in Sphinx, i.e. recreate the theory manual :-). Really I just wanted to use this pull request as a proposition of a way we can handle the documentation going forwared. We don't have to abandon NaluDoc if you would rather have it as a separate LaTeX document, but in my opinion I think it is worth having it build on Readthedocs automatically to make it low overhead for the user. However, I think it would be great to see the developer and user documentation use this Doxygen/Sphinx framework going forward as I think it has a lot of benefits. From an organizational standpoint, I don't like the documentation hidden in a directory in the code, but I also like the idea of thinking of the documentation in the same way as the code so that it stays alive in the project, and I think that is both good for the developers and users to keep it up-to-date and easy to utilize.

In the end, the theory manual you see on Readthedocs with source located in `docs/source/theory` are merely an example of what we can move to in that case if you are interested in doing so, with the future option of getting a PDF out of the Sphinx version. I think duplicating both the Sphinx theory manual and NaluDoc is not what either of us would want to do. So ultimately I would like to know if you and others find it feasible and beneficial to use this documentation framework going forward? I think it can be whatever we want it to be, i.e. we could commit to just the Doxygen part, or Doxygen with a Sphinx user manual, or everything in Sphinx, or I can look into other options too. Either way I'm just hoping to discuss it because I think documentation means different things to different people.

michaelasprague:
I would advocate that we transition to the readthedocs venue for all documentation and deprecate NaluDoc, but I think we should have that discussion.  I like the idea of the documentation residing with the code, and hope that new code features would be then reflected in the documentation, and documentation versioning would be tied directly to code versoning.

spdomin:
I agree on the deprecation of NaluDoc. My goal is to always minimize duplication of artifacts. However, if we do move everything from NaluDoc to Nalu, we still need the ability to create a pdf in a format as shown in the theory and verification manual for R&A reasons.

michaelasprague:
The pdf pathway is straightforward as I under stand sphinx -- the documentation can easily go to either the web or latex->pdf.

spdomin:
The syntax seems to be very different, e.g.,
```
.. math::
   :label: phi-central

   \phi_{cds} = \sum N^{ip}_k \phi_k.

```
Sometimes, theory goes to papers - which are written in latex. Is there an ability to do all of this within latex?

jrood-nrel:
@gantech Might be able to provide some insight here. In my experience it took some human intervention when trying to convert latex to sphinx, so I'm not sure there is a truly automated way to accomplish that.

jrood-nrel:
Actually when you say that the theory goes to papers which are written in latex, do you mean you are copying and pasting from papers you have written in latex? [Pandoc](http://pandoc.org) is the main tool I have been using which can convert between latex and sphinx which might help but probably has a lot of quirks I don't know about. My next question would be how dynamic the theory manual content would be? If it's going to change a lot and latex is preferred, I am fine omitting it from Readthedocs.

mbarone81:
Hi all.  I like the idea of a comprehensive and automated doc system that lives with the code.  Full support for LaTeX->PDF should be a requirement wrt/ the theory and verification manuals. @gantech, what type of intervention is required?  Is the problem that this feature of Sphinx is incomplete?

gantech:
Hello.. sorry for the delayed response. The conversion to latex is not straightforward right now. I am doing some manual interventions there. However, it is certainly possible to automate this. The path to a pdf document is indeed rst -> latex -> pdf. I'll send you all an example of a pdf document I have been able to generate so far. I'll then work with @jrood-nrel to show current issues with complete automation of the latex + pdf generation process.

PR #69, opened 2017-02-09
Add a supplemental algorithm builder function

rcknaus:
Cleans up some of the logic around building the new templated supplemental algorithm.

* Need to add a "name" to the supplemental algorithm to use the builder function.  The name is what the user will specify in the input file to activate the supplemental algorithm.

* Logs the name of whatever supplemental algorithm that is built (or might be built) by the builder.  So we can output a list of available supplemental algorithms for a particular physics type.

### DIFF
--- a/include/AlgTraits.h
+++ b/include/AlgTraits.h
@@ -9,6 +9,8 @@
 #ifndef AlgTraits_h
 #define AlgTraits_h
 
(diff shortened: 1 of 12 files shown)


spdomin:
This implementation is not working with my hybrid case. Therefore, I will revert.

PR #70, opened 2017-02-09
Revert "Add a supplemental algorithm builder function"

spdomin:
Reverts NaluCFD/Nalu#69


spdomin:
@rcknaus, I am in the process of modifying the cvfemHC to use the hybrid cylinder mesh. When this is passing, feel free to issue a new pull request.

PR #72, opened 2017-02-10
Changing nightly testing code to use new testing script at NREL that builds dependencies with Spack.

jrood-nrel:
This change allows me to use a new nightly testing script (https://github.com/jrood-nrel/NaluSpack/blob/master/test_nalu_peregrine.sh) that will build and test a 2x2 matrix of combinations (GCC, Intel, Trilinos master/develop) that uses Spack underneath.


spdomin:
Looks great. I think that if we have a "recent" flavor of Intel, Gcc and MacX, that would be great.

There is also a question on debug. What are your thoughts?

PR #76, opened 2017-02-22
Incremental improvement to supp-alg data-sharing mechanism.

alanw0:
1. No longer hold 3 arrays of views, each of length num-fields.
   Now just hold 1 array of views. Still of length num-fields for now.

2. Change ElemDataRequests::add_gathered_nodal_field to now take
  'scalarsPerNode' in addition to the field.
   When scalarsPerNode is 1, it will create a 1D scratch-view of length
   nodes-per-element. When scalarsPerNode is greater than 1, it will
   create a 2D scratch-view with dimensions (nodes-per-element , scalarsPerNode).

   This change allows for error-checking to make sure a given gathered
   field is requested with consistent dimensions.

   This change also allows us to avoid internally calling the
   stk::mesh::FieldBase::max_size method.

   The functions for retrieving the scratch views are still named
   get_scratch_view_1D/2D/3D/4D. The names must be different because
   the return-types are different types and you can't overload on
   return-type alone.

3. Don't need to store element-nodes in a view, and pointer is
   sufficient. (May not even need element-nodes on shared-data, but
   we can further evaluate that later.)

4. Remove un-needed 'add' functions on ElemDataRequests. We are
   using the more descriptive 'add_gathered_nodal_field' and
   'add_master_element_call', don't need a generic 'add'.

5. Remove the un-needed 'NODES' enum from ElemDataRequests.

Profiled the hybrid-cylinder case to ensure that performance is still
good. There may be an opportunity to improve the performance of the
gather through templating.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -52,6 +52,17 @@ void gather_elem_node_field(const stk::mesh::FieldBase& field,
   }
 }
 
+struct ViewHolder {
+  virtual ~ViewHolder() {}
+};
+
+template<typename T>
+struct ViewT : public ViewHolder {
+  ViewT(T view) : view_(view) {}
+  virtual ~ViewT(){}
+  T view_;
+};
+
 class ScratchViews
 {
 public:
@@ -59,7 +70,7 @@ public:
                const stk::mesh::BulkData& bulkData,
                stk::topology topo,
                ElemDataRequests& dataNeeded)
-  : elemNodes(), scs_areav(), dndx(), deriv(), det_j()
+  : elemNodes(nullptr), scs_areav(), dndx(), deriv(), det_j()
   {
     // master elements are allowed to be null if they are not required
     MasterElement *meSCS = dataNeeded.get_cvfem_surface_me();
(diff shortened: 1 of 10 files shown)


include/ScratchViews.h
```diff
@@ -223,32 +248,31 @@ void fill_pre_req_data(
...
   MasterElement *meSCS = dataNeeded.get_cvfem_surface_me();
   MasterElement *meSCV = dataNeeded.get_cvfem_volume_me();
 
   const FieldSet& neededFields = dataNeeded.get_fields();
-  for(const stk::mesh::FieldBase* fieldPtr : neededFields) {
-    stk::mesh::EntityRank fieldEntityRank = fieldPtr->entity_rank();
+  for(const FieldInfo& fieldInfo : neededFields) {
+    stk::mesh::EntityRank fieldEntityRank = fieldInfo.field->entity_rank();
+    unsigned scalarsPerEntity = fieldInfo.scalarsPerEntity;
     if (fieldEntityRank==stk::topology::ELEM_RANK) {
```
spdomin: Okay, so this means that there is still an interface to elemental data that is shared?

include/ScratchViews.h
```diff
@@ -223,32 +248,31 @@ void fill_pre_req_data(
...
+      double* fieldDataPtr = static_cast<double*>(stk::mesh::field_data(*fieldInfo.field, elem));
       for(unsigned i=0; i<len; ++i) {
         shmemView(i) = fieldDataPtr[i];
       }
     }
     else if (fieldEntityRank == stk::topology::NODE_RANK) {
-      SharedMemView<double*>& shmemView1D = prereqData.get_scratch_view_1D(*fieldPtr);
-      if (shmemView1D.dimension(0) > 0) {
-        gather_elem_node_field(*fieldPtr, elem, nodesPerElem, bulkData.begin_nodes(elem), shmemView1D);
+      if (scalarsPerEntity == 1) {
```
spdomin: This works now for scalar and vectors (sized by nDim). I think that in some rare cases, we have tensor quantities.

alanw0:
Stefan, I just added a second commit which provides for tensor fields in the data-sharing mechanism, for both gathered-nodal and element fields.

alanw0:
@spdomin P.S.: once you accept this, I would like to take a little more time and add better unit-testing and refactor the ElemDataRequests and ScratchViews code. I think the code has outgrown those headers now, need to move some code to .C files, and also the unit-testing needs to be improved. There is too much un-tested code in there.

alanw0:
@spdomin I pushed another commit here, with good unit-testing! Also some cleanup of the headers. I think this revision is good to go now, but we can still talk about further desired features and/or other issues.

spdomin:
Yes, looks great. I will deploy a fluids solve to use the new interface. I will likely need to add a few new master element calls, however, I can manage this.

PR #81, opened 2017-03-02
HOCVFEM: rewrite promotion using stk edges, faces and misc. changes

rcknaus:
Still don't have the verification testing automated yet.  I'll work on getting that set, but I wanted to open the pull request for review, specifically element_promotion/PromoteElement.C

* Need to change Trilinos build: add -DTrilinos_ENABLE_STKUnit_test_utils:BOOL=ON
* New, lighter weight ElementDescription based on edges/faces for connectivity.
* New element promotion routine, relying on stk's edge/face creation methods to determine node sharing and permutation relationships
* Added tests for the element description maps
* Converted some unit tests for the old element promotion routine to the new one
* Switching to use 64-bit in the IOSS properties meant that Visit no longer works for visualizing the high order meshes.  Paraview seems to work fine.

Doesn't combine with the old hoCVFEM branch unfortunately


spdomin:
Thus far, this is looking very nice. I would like to have more input on this code change. Let's have at least three people on the @NaluCFD/core team that understand this methodology. Any volunteers?

gantech:
Hello.. I'd volunteer to understand this. I'm supposed to work with Robert on the the higher order stuff. Do I need to understand this before we accept the pull request?

spdomin:
@gantech, well, not completely. This part of the code is nitty gritty STK details and I would like more than two people to be on the same page on how it all works. I am not sure who is the best third person. If you plan on being more than a user of the higher order work, then you would be a prime candidate to work through the implementation.

spdomin:
I am getting close to accepting this (been through the entire code). Are we sure that all the tests (performance and nightly) pass? The only concern is the fact that we are removing face node relations calls in favor of a new master element call. I want to make sure that this is correct. However, a unit test might be best since I am sure that the rtest suite is not guaranteed to contain element ordinals for the superset of all possible options.

rcknaus:
Both the nightly and perf regression tests all passed.  For unit testing the new master element call, I just have  [UnitTestSideNodeOrdinals.C](https://github.com/rcknaus/Nalu/blob/promotion_refactor/unit_tests/UnitTestSideNodeOrdinals.C) that checks that the new calls produce the same result as going through stk topology.

spdomin:
@rcknaus, please check to make sure that debug builds are clean. Thanks and good work.

rcknaus:
@NaluCFD/core 

Just to make sure everyone catches this: the trilinos config needs to be updated with a line: 

 -DTrilinos_ENABLE_STKUnit_test_utils:BOOL=ON

PR #85, opened 2017-03-07
Node balancing patch applied and modified to output individual proces…

mbarone81:
…sor node counts.


spdomin:
You should probably work this into the adaptivity use case. Specifically, if (adapt) doNodalBalance(). This may be hard to test since we do not have adaptivity active in Nalu. Otherwise, code looks good. I still prefer to see this in STK sooner than later.

spdomin:
@mbarone81, could you add a regression test coverage for this new feature? Possibly the hybrid heat conduction simulation might be good, cvfemHC.

PR #90, opened 2017-03-08
Updating CTest to be more modular.

jrood-nrel:
This pull request has some reorganization to the CTest tests. This is mostly something Shreyas and I have wanted to have implemented. The summary is that it allows the developer to enable CTest to run the regression tests locally. This same list of tests is what is run nightly. This means the developer can enable the tests at configure time by specifying the 'NaluRtest' and 'runNaluRtest' directories and use 'make test' to run the tests. The power of this is that the developer can then use CTest's ability to run subsets of tests, e.g. something like 'make femHC' to run that single test, and more such as regular expression matching.

The list of changes is as such: There is a minor addition to the main CMakeLists at the bottom. The rest is confined to moving files from the `cmake` directory involving testing to a `reg_tests` directory. Then I have a single list of the tests in `CTestList.cmake`. Whenever the tests are run, the `CTestPrepareTests.cmake` script is run (by use of a pre-test command in `CTestCustom.cmake`) to prepare `runNaluRtest` for each test.


sayerhs:
@jrood-nrel This looks good! I will try fetching and testing this out.

jrood-nrel:
@sayerhs There are so many moving parts it's difficult and slow to test the tests, but it's mostly there and I've tested `make test` and the nightly tests with only a few tests enabled and it worked as expected. I kind of need to see it in the official repo to be able to finish it off and clean it up though.

alanw0:
I haven’t looked at this in detail but it sounds great!
I’m looking forward to using it.

Alan

spdomin:
Yes, this is something that I have wanted from day one. What is the plan to phase out our manual scripts?

jrood-nrel:
Thanks for merging this. My plan is to continue trying to get to parity with the manual scripts before fully committing to CTest in the future and after nailing down the agreed upon configuration and tolerances for a gold standard here. At some point I may have to mess around in NaluRtest to fix some non-uniformity to get all of the tests going in my framework. I have also had thoughts about deprecating `runNaluRtest` in the future and maybe running directly in 'NaluRtest' with relative paths to meshes or something? I haven't thought about it much. By that point though, I will want to be quite comfortable with CTest. So the phase out plan would be whenever we're comfortable with the functionality in CTest, we can start to abandon the test scripts. I still need to enable some tests and iron out tolerance per test for full functionality.

PR #92, opened 2017-03-09
Add some basic unit tests for the master element routines

rcknaus:
* Add SCV ips / shape functions for tris
* Pyramids don't pass the patch tests.  I'd have to specialize them for pyramids.
* Still missing isInElement / interpolatePoint methods for HO


spdomin:
What is the problem with pyramids on a simple shape function evaluation? My specific interest is making sure that the consistent and lumped mass time terms work for all current topos. Is this not the case?

PR #93, opened 2017-03-14
Update to actuator line implementation with FAST coupling

gantech:
@michaelasprague  Actuator Line coupled to FAST for Q1 deliverable


spdomin:
Let's add @rcknaus and @spdomin to the review. If readability is increased by squashing, feel free to do this. Also, feel free to make updates to this pull request as you clean the code up.

gantech:
Matt C and I will continue to make changes in our fork to clean up some code. We will squash the commits after we're done with all the changes.

spdomin:
What is the current status of this capability? Has anyone run a real A/L simulation?

I have some tests in line that would evaluate the current ghosting approach. Thus far, the cost of custom ghosting and communicating seems small in my spot cases. However, I have a really interesting explicit LES filter approach that will really push the limits of the current parallel design.

gantech:
I'm testing the second approach of ghosting the actuator points to the processors. I'm also working on perform a timing study on a real case with both approaches. I hope to be done with it all by this weekend.

gantech:
I have tested different ghosting approaches and made a lot more changes. Please don't consider any of this final yet.

spdomin:
@gantech, what was the speed difference between the original custom ghosting approach (elements within a given radius ghosted to the point in which the A/L proc resides) vs ghosting data to the off proc elements within a given radius?

gantech:
@spdomin I compared the two ghosting approaches on a two turbine case with 8M cells running on 168 processors for 100 time steps. 

Ghost Elements - The existing setup - elements within a given radius ghosted to the point in which the A/L proc resides
Ghost Actuator points -  Ghosting point data to the off proc elements within a given radius

The ghosting element approach wins over the ghosting actuator points approach - at least the way I implemented it.

compareghost_elements_actuatorpoints

p.s: I also compared the two approaches on a 64M mesh with 1344 processors with similar results.

spdomin:
That is amazing - although confirms my experiments using this ghosted elements approach for an explicit LES-type filtering.

My first thought is that I am suspicious of the ghosting of actuator line points:) @sayerhs, what are your thoughts?

gantech:
So I did discuss my implementation of the ghosting of actuator points with @sayerhs and @rcknaus. It's in the ghostActPts branch of my fork here https://github.com/gantech/Nalu/blob/ghostActPts/src/ActuatorLineFAST.C

I suspect that the MPI transfer of the actuator line points and the forces at those points to the related processors and getting the sampled velocity back to the processor owning FAST is taking up quite a bit more time than I expected. My implementation of the ghosting of actuator points not only performs poorly compared to the ghosting of elements for a given number of turbines, it also seems to scale worse than the ghosting elements approach with increasing number of turbines. I expected the opposite results for both.

spdomin:
Your choice on how to proceed. However, as noted, I have tested this parallel pattern using explicit LES filtering. In this case, every point/proc in the domain has a collection of elements ghosted to it. Moreover, I added a fake integration to explore the performance of ghosted->to-shared/owned parallel assembly and the timings were also scalable.

spdomin:
@gantech 

Note that I just removed the Octree search in my last push. You will need to do the same in your A/L implementation so that STK can deprecate this search.

https://github.com/NaluCFD/Nalu/commit/20ac28b3f5591e3968477acac30b2e8319d6b096

gantech:
Ok.. I see that you updated the ActuatorLine.C in the commit as well. I'll change my fork accordingly. Thanks.

michaelasprague:
@gantech @spdomin: What are the remaining issues that need to be addressed before this pull request can be merged?

gantech:
I'm doing one final cleanup with @sayerhs to remove unwanted code and fix existing actuatorLine test to work with the modified setup (no change in gold files). I'll make another push to the NaluMesh repo to add the files associated with nrel5MWactuatorLine regression test.

gantech:
My update is complete. All tests pass. The NaluMesh repo has been updated to include the nrel5MWactuatorLine test files. The existing actuatorLine test input deck was modified to work with the new actuator class hierarchy. I think it's ready to be merged into master.

spdomin:
I will take a look and report back. Most times I have looked at the pull
request, the code has been in a state of "conflict". It looks clean now
after the last set of commits. I suppose now we can go over the final
design aspects of the effort. With any hope the state will be clean and
ready to push.

spdomin:
Let's talk about this code at the next standup. I guess at first blush I do not understand why all of the recent code changes, e.g., mdotAlgDriver, DixedPressure. Etc. are showing up as changes G made?

sayerhs:
@spdomin Looking at the _commit tree_ (see attached image) and running `git blame` on the file you mention ([ComputeMdotAlgorithmDriver](https://github.com/NaluCFD/Nalu/blame/b8e0101095e9ff50d01bc7b894f052c53157caea/include/ComputeMdotAlgorithmDriver.h)) it appears that authorships of the files haven't been changed. It still shows you as the author of the files and preserves the original commit that introduced the files. 


screen shot 2017-08-08 at 7 40 03 am

screen shot 2017-08-08 at 7 42 57 am

screen shot 2017-08-08 at 7 44 20 am

sayerhs:
@spdomin And what the git history would look like after merging this pull request to `NaluCFD/Nalu/master` branch (i.e., no squash merge). This seems like normal git behavior to me. 

screen shot 2017-08-08 at 7 53 59 am

spdomin:
@sayerhs, good to know. Often times, I work pull requests through GH. In this case, I will clone and work through the changes. Thanks.

build/do-configNalu_release
```diff
@@ -14,6 +15,7 @@ cmake \
   -DYAML_DIR:PATH=$yaml_install_dir \
   -DENABLE_INSTALL:BOOL=OFF \
   -DCMAKE_BUILD_TYPE=RELEASE \
+  -DOpenFAST_DIR:PATH=openfast_install_dir \
```
sayerhs: Can we also explicitly add `-DNALU_USES_OPENFAST:BOOL=ON` here, so that users can disable that if they are not prepared to install OpenFAST as yet.

CMakeLists.txt
```diff
@@ -11,6 +10,8 @@ cmake_minimum_required (VERSION 2.8.11)
 #
 #  THESE PATHS MAY NEED TO BE CHANGED DEPENDING ON WHERE THEY ARE STORED!!!
 #
+option(NALU_USES_OPENFAST
```
sayerhs: Need to coordinate with @jrood-nrel on the nightly testing for merging this pull request.

docs/source/theory/actuatorLine.rst
```diff
@@ -0,0 +1,185 @@
...
+:eq:`force-integral` so long as the ratio of actuator element size to
+projection function width :math:`\epsilon` does not exceed a certain threshold.
+
+Design
+======
+
+The initial actuator capability implemented in Nalu is focused on the actuator line algorithm. However, the class hierarchy is designed with placeholders for actuator disk and swept actuator line capability and potentially an actuator surface capability. The general set of functions required to handle actuators are the same whether it is a line, disk, swept line, or surface. The various classes implementing the different actuator capabilities inherit from an overall actuator class. The ``ActuatorLineFAST`` class couples Nalu with the third party library OpenFAST for actuator line simulations of wind turbines. OpenFAST (https://nwtc.nrel.gov/FAST), available from https://github.com/OpenFAST/openfast, is a aero-hydro-servo-elastic tool to model wind turbine developed by the National Renewable Energy Laboratory (NREL). The ``ActuatorLineFAST`` class will help Nalu effectively act as an inflow module to OpenFAST by supplying the velocity field information.
```
sayerhs: Removed the placeholder classes in the latest commits. Should this be rephrased?

src/Realm.C
```diff
@@ -545,13 +550,37 @@ Realm::look_ahead_and_creation(const YAML::Node & node)
...
+      switch ( ActuatorTypeMap[ActuatorTypeName] ) {
+      case ActuatorType::ActLinePointDrag : {
+	actuator_ =  new ActuatorLinePointDrag(*this, *foundActuator[0]);
+	break;
+      }
+#ifdef NALU_USES_OPENFAST
+      case ActuatorType::ActLineFAST : {
+	actuator_ =  new ActuatorLineFAST(*this, *foundActuator[0]);
+	break;
+      }
```
sayerhs: Add `#else` here to warn users that OpenFAST integration was requested but was not enabled during compile time?

gantech:
Addressed latest comments.

spdomin:
Although I have not performed a bisect, it looks like the actuator line test case failed with appreciable diffs. However, this test should have been orthogonal to the FAST coupling as it represents the former Stokes-drag approach.

Can we have someone look at this or, alternatively, revert the commit?

gantech:
That's weird. It passed when I ran the tests. I'm checking this.

spdomin:
I double checked and also confirmed this with the dash. A bisect on your commit might be the fastest way to figure out the issue?

gantech:
The diff is on the second to last commit "Merge branch 'master' into splitActuatorLineImplementations" - d0cbd89. The commit before that passes the test. Continuing to work to identify the problem.

gantech:
I'm sorry for this. My latest merge caused diffs in a file where I did not expect it. I fixed the issue and all actuatorLine tests pass now with the same gold standard files.

spdomin:
Understood. I think Shreyas noted the subtlety of the pull request merge above and also spoke to the need for the review to clone the project and work through gitk to identify real vs phantom changes that appeared though the GH interface.

Fell free to either push directly or issue another pull request. If you can keep an eye on the dash the next day or so, that would be great.

PR #94, opened 2017-03-14
Fixed minor typo and changed output precision width.

marchdf:
`Fxz` should be `Fvz` and the output precision width was causing columns of data to fuse
together, leading to something like the attached screenshot.
screen shot 2017-03-14 at 12 09 02 pm

This may have to be changed in other files? Maybe we should consider a "global" value for this?


spdomin:
Looks good. I think that there are not many places where such a width is specified. Possibly solutionNorms, and line-of-site, however, I have not seen a clash in size yet..

Thanks for finding this.

PR #99, opened 2017-03-17
Implementation of the Moeng surface shear stress boundary condition for

mbarone81:
simulation of atmospheric flows.  This boundary condition is activated by
specifying 'use_moeng_wall_function' in the wall_user_data section of the
input file.  This commit contains only the shear stress boundary term
for the momentum system.  For non-isothermal runs an enthalpy boundary
flux is required as well.

Fixed bug with assigning proper ABL velocity profile function when surface temperature flux was zero.  Affects both ABL wall function and Moeng wall function.

Added algorithm for computing forces and moments for the Moeng ABL wall function.
Since the Moeng condition does not utilize a local friction velocity, a surface field
was created to store the local surface shear stress at the face integration points.  The
surface force and moment algorithm assembles these to the nodal tau_wall field.

Modified the BoundaryLayerPeturbationAuxFunction class to accept a U and V
velocity input to allow for boundary layer initialization for arbitrary flow
direction in the x-y plane.

### DIFF
--- a/include/AssembleMomentumMoengWallFunctionSolverAlgorithm.h
+++ b/include/AssembleMomentumMoengWallFunctionSolverAlgorithm.h
@@ -1,0 +1,72 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+
+#ifndef AssembleMomentumMoengWallFunctionSolverAlgorithm_h
+#define AssembleMomentumMoengWallFunctionSolverAlgorithm_h
+
+#include<SolverAlgorithm.h>
+#include<FieldTypeDef.h>
+#include<ABLProfileFunction.h>
+
+namespace stk {
+namespace mesh {
+class Part;
+}
+}
+
+namespace sierra{
+namespace nalu{
+
+class Realm;
+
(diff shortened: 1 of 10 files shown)


mbarone81:
Please consider this pull request for review only at the moment.  It will break one of the ABL nightly tests because of a bug fix in the ABLWallFunction boundary condition.  I will submit a modified Rtest shortly.

spdomin:
Is this ready to go then? The comment above still suggests that the effort is a WIP.

Also, in general, the code should allow for mixing and matching of all types of wall bcs.

mbarone81:
@spdomin many thanks for the comments.  As I review the implementation, I see a problem that if I have registered an abl_wall_function for example, then if I then try to register a moeng_wall_function, that will not be successful because both of these are associated with wfAlgType=WALL_ABL.  A solution is to provide a unique wfAlgType for each bc algorithm.  Does this address the issue(s) in your review?

spdomin:
I think so. There is also a problem with mixing and matching dirichlet bcs that we should resolve using the standard wall function approach. We should sit down together to fix that. I am not sure if that was a problem before or something you modified? However, the intent of bcs is clear in that we should be able to mix and match things arbitrarily. I had a case once that wanted a wall function on an upper experimental chamber with resolved flow on an object...

mbarone81:
I created some new enums for the wall bc algTypes that should address the mixing and matching of different wall BCs.  Closing this pull request and will submit a new one.

spdomin:
@mbarone81, please take a look at the standard wall function approach to ensure that this model will also plug and play. Otherwise, the code push looks fine. As noted, this will need to be re-factored when we deploy the consolidated approach with templating.

PR #100, opened 2017-03-21
Possibility of suppressing deprecated declaration warnings until a new Yaml-cpp release exists

jrood-nrel:
This is a purely cosmetic pull request I'm opening for consideration. Nalu has warnings due to Yaml-cpp 0.5.3 using `auto_ptr` which is considered deprecated. Looking at the current Yaml github, they have moved onto using `unique_ptr` which should resolve these warnings in a future Yaml-cpp release. In the meantime, we can have the luxury of ignoring these deprecated declaration warnings, unless I'm missing a good reason not to.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,12 +79,14 @@ ENDIF(Trilinos_BUILD_SHARED_LIBS)
 
 PROJECT(Nalu)
 
+# Suppress deprecated declarations until next Yaml-cpp
+# release uses unique_ptr instead of the deprecated auto_ptr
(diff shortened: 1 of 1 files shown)


jrood-nrel:
@spdomin No problem. It was just a thought, and I'll close this. Unfortunately YAML hasn't tagged a new version. I was just looking at the master branch. I think we should wait until they tag a new version to move to it.

spdomin:
let's submit an issue then to see if they can tag and release?

jrood-nrel:
@spdomin So it looks like there is an issue for this listed: https://github.com/jbeder/yaml-cpp/issues/175 .

Issue #13, opened 2016-10-10
Normal distance to the wall usage for resolved and wall function simulation


spdomin:
When computing y+, what is the value of Yp that should be used?

The current wall-resolved implementation is using a normal distance = 1/4*normalOpposingEdgeDistance. However, based on the usage of the projected nodal gradient, this distance should simply be normalOpposingEdgeDistance.

For the wall-function approach, more thought is required.

spdomin:
Wall-resolved case improved: commit 16e309a82244a5c626303c195e48847de5766c67

Higher order may add complexity - or not...

spdomin:
@mbarone81, while you are working on the ABL, think about the wall-modeled approach and the appropriate normal distance to use. I am thinking that it is also the full edge-length.

mbarone81:
Will do -- my first thoughts are that full edge length is appropriate.

spdomin:
Recent LOW findings suggest that the projected approach is the most ideal implementation when using wall functions. 

At present, the resolved YPlus calculation is based on the full edge-length and the Yplus for standard wall functions is 1/4 edge distance.

PR #101, opened 2017-03-22
Adding SDRWallFactor as a parameter so BC can be matched to NASA.

marchdf:
NASA's description of the BC for the Menter SST model includes a wall
factor of 10. The wall factor is now exposed to the user as a
parameter.

### DIFF
--- a/include/Enums.h
+++ b/include/Enums.h
@@ -171,7 +171,8 @@ enum TurbulenceModelConstant {
   TM_cmuCs = 17,
   TM_Cw = 18,
   TM_CbTwo = 19,
-  TM_END = 20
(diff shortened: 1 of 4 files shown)


spdomin:
This looks fine. Note that the manner in which the normal distance to the wall is chosen generally guides what value this parameter uses. Feel free to look at the code and choose this parameter appropriately based on the current Yp calculation.

If we plan on driving a RANS production capability in Nalu, which I think the project is driving, we should review the following:

https://github.com/NaluCFD/Nalu/issues/13

- and even consider some canonical backward facing step simulations in addition to some MMS for SST.

PR #105, opened 2017-04-04
Remove factor of 2 in numBytes calculation for ScratchViews

sayerhs:
Fix segfault by adding 64 extra bytes to account for any alignment
requirements.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -60,6 +60,8 @@ public:
   inline
   SharedMemView<double****>& get_scratch_view_4D(const stk::mesh::FieldBase& field);
 
+  inline const int total_bytes() const { return num_bytes_required; }
+
   const stk::mesh::Entity* elemNodes;
   SharedMemView<double**> scs_areav;
   SharedMemView<double***> dndx;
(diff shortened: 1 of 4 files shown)


sayerhs:
@alanw0 I see the logic in adding the padding to the `get_num_bytes_pre_req_data` function. However, the total bytes has the `rhsSize + lhsSize + scratchIdsSize` which is calculated in AssembleElemSolverAlgorithm. Do you think those sizes should also be consolidated in the function? 

I'll add some updates to UnitTests to check for sizes. But to catch segfaults we need access every element of the array, just `EXPECT_EQ` won't work. Do you have a better suggestion for that?

alanw0:
@sayerhs the only reason I think it makes sense to keep rhsSize and lhsSize outside of get_num_bytes... is because the rhs and lhs arrays aren't part of the pre-req data (ScratchViews).

Regarding the unit-tests, I think that if an EXPECT_EQ can confirm that the amount of memory used within ScratchViews is less than the value returned by get_num_bytes (and add in the padding as appropriate), then that at least provides some protection against get_num_bytes getting out of sync with what ScratchViews is requesting.

sayerhs:
@alanw0 Please take a look at the changes. 

- Moved padding to `get_num_bytes_pre_req_data` 
- Added an `EXPECT_EQ` to ensure that the `ScratchViews` data requirement matches what we estimate in `get_num_bytes_pre_req_data`.

PR #107, opened 2017-04-10
modify getDofStatus to work with both noncormal and periodic.

skennon10:
* add part attributes to identify non-conformal and periodic boundaries

   * modify getDofStatus to use part attributes instead of global Realm flags

   * note: small tests with periodic and MueLu only work if MueLu parameter

     <Parameter        name="coarse: max size"                 type="int"      value="100"/>

     is set small enough to force a coarse solve (thanks Stefan for this suggestion)"

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -79,9 +79,6 @@ class LagrangeBasis;
 class PromotedElementIO;
 struct ElementDescription;
 
-struct PeriodicPartAttribute {};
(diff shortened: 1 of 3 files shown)


spdomin:
@skennon10, I will pull this down and test ASAP given my weeks work load.

Also, I never receive and email for such requests. Specifically, I need to log into GH to see what my unread notifications are.

We will probably want @alanw0 to have a look as well.

PR #108, opened 2017-04-11
incorporated review comments

skennon10:


src/TpetraLinearSystem.C
```diff
@@ -155,28 +155,28 @@ int TpetraLinearSystem::getDofStatus(stk::mesh::Entity node)
   bool has_non_matching_boundary_face_alg = false;
   bool hasPeriodic = false;
 
-  bool debug = false;
+  if (NULL == realm_.allPeriodicInteractingSelector_ && realm_.allPeriodicInteractingParts_.size())
```
spdomin: I do not think that a selector needs to be a private data member as STK will do this for us behind the scenes. I think we simply need to store the partVector for each.

skennon10: I used a Selector that is created once for efficiency reasons - I just assumed that selectUnion was expensive, and this method gets called for every node.  It would be cleaner to only store the PartVector, however.

alanw0:
I agree with @spdomin that it would be good to check performance just to be sure we know. Sometimes a performance hit is necessary when fixing a correctness issue... But of course it's good to maintain performance whenever possible. I have the cube6M files on Serrano, I'd be happy to provide @skennon10 with queue submission scripts etc if he's not already set up to run on that platform. (Same scripts run on chama etc.)

spdomin:
Is this effort stalling? Let's touch base early next week to see what we can do to push.

spdomin:
@mbarone81, might you have any cycles to clean this code up and push? We can work together next week if you like..

mbarone81:
OK, yes, I can work on this, probably towards the end of this week/beginning of next week.

skennon10:
I have incorporated changes from Stefan and my discussion but now some tests are failing (unrelated to my changes).  What Trilinos version are we using?

spdomin:
I think that this one is closed as you have a new one with the changes suggested...

PR #111, opened 2017-04-17
Enable paraview catalyst

tjotaha:
The changes are in Realm.C and OutputInfo.C, and OutputInfo.h. The build will succeed as long as you are building against the Trilinos master branch and you set the following flag in your Trilinos build:

-DTrilinos_ENABLE_SEACASIoss:BOOL=ON

ParaView Catalyst will only be active when an "output" block in a Nalu input deck contains a "catalyst_file_name" command. It is also necessary to build the ParaView Catalyst adapter against an install of ParaView and set an environment variable to point to the install location. We will need to update the Nalu build instructions to show these details. 

The environment variable is called CATALYST_ADAPTER_INSTALL_DIR. It should be set to the root of the install directory for the Catalyst adapter.

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -84,6 +84,7 @@
 #include <stk_util/parallel/Parallel.hpp>
 #include <stk_util/environment/WallTime.hpp>
 #include <stk_util/environment/perf_util.hpp>
+#include <stk_util/environment/FileUtils.hpp>
 
 // stk_mesh/base/fem
 #include <stk_mesh/base/BulkData.hpp>
@@ -118,6 +119,9 @@
 #include <cmath>
 #include <utility>
 #include <stdint.h>
+
+// catalyst visualization output
+#include <Iovs_DatabaseIO.h>
 
 #define USE_NALU_PERFORMANCE_TESTING_CALLGRIND 0
 #if USE_NALU_PERFORMANCE_TESTING_CALLGRIND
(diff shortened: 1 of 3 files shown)


src/Realm.C
```diff
@@ -1928,7 +1942,46 @@ Realm::create_output_mesh()
       if (fileid++ > 0) oname += "-s" + fileid_ss.str();
     }
 
-    resultsFileIndex_ = ioBroker_->create_output_mesh( oname, stk::io::WRITE_RESULTS, *outputInfo_->outputPropertyManager_);
+
+    if(!outputInfo_->catalystFileName_.empty()||
```
spdomin: Is the concept here that one obtains either Catalyst **or** standard output? Is there a manner to support both in a single simulation? I think that would be most preferable.

tjotaha: Yes, the implementation either writes to Catalyst or to standard output but not to both. In Sierra, we duplicate a results output block in the input deck, and have one block for Catalyst and one block for Exodus II output, if desired. In Nalu you can only have one output section in the input deck?

spdomin: In theory, we should be able to have multiple blocks without too much issue. For now, we can move with the either/or path and clean up the I/O to allow for a general amount of output blocks. At present, we support one unique output and restart block.

spdomin:
@tjotaha, can you please add a pull request to modify the Trilinos configure file?

PR #117, opened 2017-04-20
CTest reorganization

jrood-nrel:
This is the large pull request that pulls NaluRtest into the main repo while keeping meshes external using a submodule. Tests here run in the build directory and no file copies are necessary.


jrood-nrel:
Yeah I kind of wanted to set an example by adding the documentation in with the pull request, but realized I should wait until I have this running nightly and have ironed out any bugs with it before adding the documentation in case I have to change something. That will come soon after though, but not sure if it will be before or after I move to Merlin.

spdomin:
I test nightly and will need to know what to do before I merge this in...

jrood-nrel:
@spdomin 

Sorry, I am not intending that this flips the switch just yet. This actually doesn't modify anything involved with your current nightly testing. This will only affect NREL at the moment, but in a big way. I would like to merge it because I would like to test this new testing reorganization and I will have to sync up some external factors on Peregrine to do so after this merge.

NaluRtest should still work as normal while I help get you a nightly test script where your testing uses CTest, in which case, I think we should still have your NaluRtest nightly tests as a backup until we are getting consistent results with a new script that you are comfortable with on your end.

Can you point me to your nightly testing script? Is it the one in the repo already (under `nightly_build`)? I can help convert it directly first, or I can start on documentation that describes how to add a test machine and we can work off of that.

Does that sound ok?

PR #125, opened 2017-05-04
incorporate performance-related review comments

skennon10:


src/TpetraLinearSystem.C
```diff
@@ -152,11 +152,37 @@ int TpetraLinearSystem::getDofStatus(stk::mesh::Entity node)
   const bool entityIsShared = b.shared();
   const bool entityIsGhosted = !entityIsOwned && !entityIsShared;
 
-  if (realm_.hasPeriodic_ && realm_.has_non_matching_boundary_face_alg())
-    throw std::logic_error("not ready for primetime to compbine periodic and non-matching algorithm suite");
+  bool has_non_matching_boundary_face_alg = realm_.has_non_matching_boundary_face_alg();
```
spdomin: Sorry, I do not understand this. If I have periodic and sliding mesh active, however, without a node touching both surfaces, then the local hasPeriodic and has_non will be false. That will cause some logic below to be missed?

PR #126, opened 2017-05-09
Kokkos ScratchViews interface update

sayerhs:
- Refactor Scratchviews to handle current and model coordinates and
      the relationship with master element data.

    - Remove coordinates from fill_pre_req_data and use appropriate
      coordinates registered by the user to populate MasterElement view
      arrays.

    - Extend master element data registration and retrieval methods to
      indicate the coordinates field used to compute the relevant data.

    - Update all Kernels and UnitTests to adhere to new ScratchViews
      interface.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -34,6 +34,34 @@ struct ViewT : public ViewHolder {
   T view_;
 };
 
+class MasterElementViews
+{
+public:
+  MasterElementViews() = default;
+  virtual ~MasterElementViews() = default;
+
+  int create_master_element_views(
+    const TeamHandleType& team,
+    const std::set<ELEM_DATA_NEEDED>& dataEnums,
+    int nDim, int nodesPerElem,
+    int numScsIp, int numScvIp);
+
+  void fill_master_element_views(
+    const std::set<ELEM_DATA_NEEDED>& dataEnums,
+    SharedMemView<double**>* coordsView,
+    MasterElement* meSCS,
+    MasterElement* meSCV);
+
+  SharedMemView<double**> scs_areav;
+  SharedMemView<double***> dndx;
+  SharedMemView<double***> dndx_shifted;
+  SharedMemView<double*> deriv;
+  SharedMemView<double*> det_j;
+  SharedMemView<double*> scv_volume;
+  SharedMemView<double***> gijUpper;
+  SharedMemView<double***> gijLower;
+};
+
 class ScratchViews
 {
 public:
@@ -60,18 +88,16 @@ public:
   inline
   SharedMemView<double****>& get_scratch_view_4D(const stk::mesh::FieldBase& field);
 
+  inline
+  MasterElementViews& get_me_views(const COORDS_TYPES cType = CURRENT_COORDINATES)
+  {
+    ThrowRequire(hasCoordField[cType] == true);
+    return meViews[cType];
+  }
+
   inline int total_bytes() const { return num_bytes_required; }
 
   const stk::mesh::Entity* elemNodes;
-  SharedMemView<double**> scs_areav;
-  SharedMemView<double***> dndx;
-  SharedMemView<double***> dndx_shifted;
-  SharedMemView<double*> deriv;
-  SharedMemView<double*> det_j;
-  SharedMemView<double*> scv_volume;
-  SharedMemView<double***> gijUpper;
-  SharedMemView<double***> gijLower;
-
 
 private:
   void create_needed_field_views(const TeamHandleType& team,
@@ -86,6 +112,8 @@ private:
 
   std::vector<ViewHolder*> fieldViews;
 
+  std::vector<MasterElementViews> meViews;
+  std::vector<bool> hasCoordField;
   int num_bytes_required{0};
 };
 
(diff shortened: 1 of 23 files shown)


include/ScratchViews.h
```diff
@@ -86,6 +112,8 @@ class ScratchViews
 
   std::vector<ViewHolder*> fieldViews;
 
+  std::vector<MasterElementViews> meViews;
+  std::vector<bool> hasCoordField;
```
spdomin: http://www.codingstandard.com/rule/17-1-1-do-not-use-stdvector/

src/ContinuityAdvElemKernel.C
```diff
@@ -102,12 +102,12 @@ ContinuityAdvElemKernel<AlgTraits>::execute(
   SharedMemView<double**>& v_velocity = scratchViews.get_scratch_view_2D(*velocityRTM_);
   SharedMemView<double**>& v_Gpdx = scratchViews.get_scratch_view_2D(*Gpdx_);
 
-  SharedMemView<double**>& v_scs_areav = scratchViews.scs_areav;
+  SharedMemView<double**>& v_scs_areav = scratchViews.get_me_views().scs_areav;
```
spdomin: This interface looks fine, however, seems more complex than what we need? Let me have more perspective on why this was changed. It may, in the end, help with BCs or FEM?

src/ContinuityMassElemKernel.C
```diff
@@ -61,7 +61,7 @@ ContinuityMassElemKernel<AlgTraits>::ContinuityMassElemKernel(
   dataPreReqs.add_cvfem_volume_me(meSCV);
 
   // fields and data
-  dataPreReqs.add_gathered_nodal_field(*coordinates_, AlgTraits::nDim_);
+  dataPreReqs.add_coordinates_field(*coordinates_, AlgTraits::nDim_);
```
spdomin: By add_coordinates_field(), you mean set the coordinates to be used for standard grad_op and areavec calcs?

As long as we support the ability to have a grad_op (based on model) and grad_op (based on current) in addition to a coordinates field used in the kernel itself, i.e., coord(nodeL) - coord(nodeR), I am fine:)

sayerhs: Yes. There is an optional parameter that is defaulted. The exact use is 

```
// Register current coordinates 
dataPreReqs.add_coordinates_field(*coordinates_, AlgTraits::nDim_, CURRENT_COORDINATES);

// Register model coordinates
dataPreReqs.add_coordinates_field(*coordinates_, AlgTraits::nDim_, MODEL_COORDINATES);

dataPreReqs.add_master_element_call(GRAD_OP, CURRENT_COORDINATES);
dataPreReqs.add_master_element_call(GRAD_OP, MODEL_COORDINATES);
```

Registers `coordinates` (both model and current) as well as `grad_op` for both these. Retrieval works like this

```
auto v_dndx_model = scratchViews.get_me_elems(MODEL_COORDINATES).dndx;
```

include/ScratchViews.h
sayerhs: This is a private field and does not require STL container behavior. So I am fine with `vector<bool>` here.

src/ContinuityMassElemKernel.C
spdomin: Oh, I missed that in the add_coordinates_field() set of calls that I looked at in the Kernel.

I will look closer at fill_master_element(0 now to see how you can push in a coordinates view and still iterate over a map of ME calcs.

src/ScratchViews.C
```diff
@@ -85,11 +85,119 @@ void gather_elem_node_field(const stk::mesh::FieldBase& field,
...
+    numScalars += numScsIp;
+  }
+
+  return numScalars;
+}
+
+void
+MasterElementViews::fill_master_element_views(
+  const std::set<ELEM_DATA_NEEDED>& dataEnums,
+  SharedMemView<double**>* coordsView,
```
spdomin: This is what I am trying to figure out... CoordsView is provided.

src/ScratchViews.C
```diff
@@ -344,44 +414,18 @@ void fill_pre_req_data(
...
-         break;
-      case SCV_VOLUME:
-         ThrowRequireMsg(meSCV != nullptr, "ERROR, meSCV needs to be non-null if SCV_VOLUME is requested.");
-         ThrowRequireMsg(coordsView != nullptr, "ERROR, coords null but SCV_VOLUME requested.");
-         meSCV->determinant(1, &((*coordsView)(0,0)), &prereqData.scv_volume(0), &error);
-         break;
-      default: break;
-    }
+
+  for (auto it = dataNeeded.get_coordinates_map().begin();
```
sayerhs: @spdomin So when the user registers coordinates (model or current), we store it in ElemDataRequests. And in the `fill_pre_req_data` call, we loop over the different `coordinates` registered and populate separate `MasterElementViews` for data related to each coordinate type.

sayerhs:
@alanw0 Updated the code based on your comments. 

`MAX_COORDS_TYPES` is getting interpreted as `constexpr`.

spdomin:
What about the std::vector<bool>? 

Also, do you think that you can modify one of the Kernles to use the full interface so that there is an example in the code base?

sayerhs:
@spdomin Please take a look at the latest commit 0770244 where I've address both these issues based on @alanw0's comments. This is part of the pull request now.

PR #130, opened 2017-05-18
Add body force element source kernel

rcknaus:
Adds a consolidate approach kernel for computing the contribution of a specified body force with a consistent mass matrix

Activate like

>        - element_source_terms:
>            momentum: [momentum_time_derivative, advection_diffusion, body_force]
>            continuity: [density_time_derivative, advection]
>
>        - source_term_parameters:
>            body_force: [1.0,0.0,0.0]

### DIFF
--- a/include/MomentumBodyForceSrcElemKernel.h
+++ b/include/MomentumBodyForceSrcElemKernel.h
@@ -1,0 +1,62 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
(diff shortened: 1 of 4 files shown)


gantech:
Awesome. I'll incorporate this into the higher order extension of the actuator line algorithm

spdomin:
@sayerhs - what is your thought on this push? Obviously, the code is fine. Wondering if we should go all in on the consolidated approach for source terms? In this case, we are integrating a constant. As such, a nodal contribution is fine.

sayerhs:
@spdomin I think we should go all in on the consolidated approach. Working with Robert will be a good way to make sure there aren't other design issues we need to address before the rest of the team start contributing to existing kernels or new kernels.

spdomin:
Yes, we agree all in on the consolidated approach. My comment was closer to this role of element and nodal contributions. Specifically, are we moving down the path that the element-based scheme only allows for element-based contributions or do we mix and match. In theory, the ElemData/Scratch will move to general node/elem/edge support..

rcknaus:
I'm going to close this.  I will probably reopen something similar in the future, though

PR #131, opened 2017-05-19
Update Kernels to use ScratchViews for element data

sayerhs:
Replace `massFlowRate` data access with ScratchViews interface in Kernel `execute` method. 

The `element` argument cannot be deprecated yet until the `FEM` kernel has been updated.


spdomin:
@sayerhs or @alanw0 - how does the element data work for hybrid meshes? I am trying to find the gather from element scalar data to the data request for mDot to make sure that the size is correct.

Thanks in advance.

sayerhs:
@spdomin: When registering the element data (in this case `massFlowRate`), we specify the size using `AlgTraits::numScsIp_` (see https://github.com/NaluCFD/Nalu/blob/master/src/MomentumAdvDiffElemKernel.C#L55). This is stored in the `FieldInfo` struct and used to populate the View in ScratchViews. Does this answer your question?

spdomin:
Okay, good to know. I think, at the very least, that we are set on the homogeneous design and such registration will always be available even if we remove templating.

PR #137, opened 2017-05-24
new test non-conformal with periodic - from Matt Barone

skennon10:


spdomin:
@jrood-nrel will need to clarify the process as we are again seeing a conflict in reg_tests/mesh?

jrood-nrel:
Yeah, oops, looks like I shouldn't have already done the mesh submodule update if it was going into the pull request. Let me try to resolve the conflict.

skennon10:
Did I do this right?  I was wondering if maybe I only needed to do a single pull request in the Nalu repo, and not a separate one in the NaluMesh repo?


________________________________

spdomin:
I think that there are two ways to proceed... The first approach is for a novice like me in that I push to NaluMesh and then update Nalu and push the test. However, in this case, you may need to iterate on the NaluMesh commits.

The second path is to modify reg_test/mesh in Nalu and perform a specific set of steps to ensure that NaluMesh is up to date.

@sayerhs suggested that we create a script that manages the second path above. 

In this case, you did in fact try to modify NaluMesh first, so I am not sure what happened. Perhaps you needed an update of sub modules on Nalu's side?

At any rate, @rcknaus had the same issue as you just did, so expert advice/tutorials on a work flow would be highly useful.

jrood-nrel:
@skennon10 You did this right. It shouldn't have had conflicts if I had I not undercut the update of the mesh submodule before this pull request, which caused the conflict. I think the approach of committing the mesh in NaluMesh first is the correct one. Then a single pull request exactly like the one that was submitted should be sufficient. Although I'm curious to know how this issue of a conflict happened to @rcknaus if so.

PR #141, opened 2017-05-31
Add NSO KE kernel test 

rcknaus:
* Adds some utilities to compute the nodal gradient of a field as well


unit_tests/kernels/UnitTestKernelUtils.C
```diff
@@ -382,6 +385,417 @@ void calc_mass_flow_rate_scs(
...
+  const auto& meta = bulk.mesh_meta_data();
+  const int ndim = meta.spatial_dimension();
+  EXPECT_EQ(ndim, 3);
+  EXPECT_EQ(topo.rank(), meta.side_rank());
+
+  sierra::nalu::ElemDataRequests dataNeeded;
+
+  auto meBC = sierra::nalu::get_surface_master_element(topo);
+
+  dataNeeded.add_cvfem_surface_me(meBC);
```
spdomin: I find this interesting... If you had a face:element algorithm, would things work out of the box? For example, if we also needed to gather interior coordinates could we do that? Probably not, right?

unit_tests/kernels/UnitTestKernelUtils.C
```diff
@@ -382,6 +385,417 @@ void calc_mass_flow_rate_scs(
...
+  const auto& buckets = bulk.get_buckets(stk::topology::ELEM_RANK, selector);
+
+  const int bytes_per_team = 0;
+  const int bytes_per_thread = sierra::nalu::get_num_bytes_pre_req_data(dataNeeded, meta.spatial_dimension()) ;
+
+  auto v_shape_function = Kokkos::View<double**>("shape_function", meSCS->numIntPoints_, meSCS->nodesPerElement_);
+
+  auto team_exec = sierra::nalu::get_team_policy(buckets.size(), bytes_per_team, bytes_per_thread);
+
+  Kokkos::parallel_for(team_exec, [&](const sierra::nalu::TeamHandleType& team) {
```
sayerhs: Can we not use `kokkos_thread_team_bucket_loop` from `UnitTestKokkosUtils.h` here? That will reduce some of the boilerplate code that we have here.

sayerhs:
@spdomin Do you mean Robert's calculation of projected nodal gradient? Looking at his code, it is hard to say how that boilerplate can be avoided. May be this design is pointing to the fact that we might need to adopt a kernel approach to refactoring some of the utilities like calculating projected nodal gradient using Kokkos views within Nalu itself?

spdomin:
Yes, that was my point. The premise of the kernel-based design was to minimize the duplicated thread-based kokkos code and STK selector/bucket approach. Although we do not yet have a kokkos non-solver algorithm approach, the concept should be the same. We could deploy this to Courant, volume, Gj(phi), etc.

In general, the extension of kokkos to non-solver algorithms is probably less critical as such timings are generally small in most simulation studies. However, I started to look at all of the Unit tests and began to worry a tiny bit on efforts to maintain them and/or change the design, such as this commit, should the kokkos interface change.

All that being said, I am fine with this push if you two sign off on it. I just wanted to express some views so that we each understand each others perspective.

sayerhs:
@spdomin If @rcknaus can see if `kokkos_thread_team_bucket_loop` is a viable approach for the loops, that would be great. If he thinks otherwise, I think we can go ahead an merge this. I'll let Robert make the decision on that.

alanw0:
I agree with the concerns about maintaining lots of boilerplate in different places, that does give the code more inertia. If the 'kokkos_thread_team_bucket_loop' can be used, that will help. But if scratch-views are involved, I don't know that kokkos_thread_team_bucket_loop is feasible. But perhaps something similar could be made to be workable.

spdomin:
I merged this now and we can clean up when the final design is deployed.

Issue #143, opened 2017-06-06
Shifted Laplace with consolidated approach is segfaulting


spdomin:
My production V27 cases are segfaulting on Cori when using the shifted element-based operators with  the consolidated algorithm approach. The segfault occurs at the ContinuityAdvectionElemKernel::execute().

In fact, the same segfault occurs on, e.g., the ductWedge case when I change the simulation to element-based, with consolidated and shifted. The modifications to the ductWedge input file to activate the shifted/consolidated approach are attached. Removal of the consolidated keyword results in a clean simulation.

I also converted the elemHybridFluids case and the same behavior is noted.

I am not sure if this was done at the time @sayerhs converted the Cont SupAlg or during subsequent refactors of current/coords or an issue with the orig SuppAlg? At the very least, the full AssembleContinuityElemSolver is working. This issue should be addressed ASAP since milestones are affected, e.g., FY17/Q3 and FY18/Q1.

[ductWedgeShiftC.txt](https://github.com/NaluCFD/Nalu/files/1055026/ductWedgeShiftC.txt)

spdomin:
Adding @mbarone81 @michaelasprague @marchdf - not to actually assign you the work - but rather to be aware of this possible bug in shifted/consolidated approaches. I think that the first input file I saw from Matt does not activate this combination, however, the plan was to include them at some point.

spdomin:
Here is the issue:
```

Options review: 
shiftMdot: no
shiftPoisson: yes
reducedSensitivities: yes

Registration phase:
shape_fcn reghistered
SCS_SHIFTED_GRAD_OP, CURRENT_COORDINATES

Execution phase 
v_dndx is using: dndx_shifted
v_dndx_lhs is using: dndx
```

spdomin:
This is my take on how the logic should read..
[shiftedLogic.txt](https://github.com/NaluCFD/Nalu/files/1055274/shiftedLogic.txt)

PR #144, opened 2017-06-06
Fix segfault in Continuity Advection Kernel

sayerhs:
- The logic for `shift_cvfem_poisson = true` was not reflected correctly
in advection kernel and caused a segfault accessing ME views that
weren't registered in construction phase.

- Added a unit test to test this flag. 

Proposed fix for #143.

### DIFF
--- a/src/ContinuityAdvElemKernel.C
+++ b/src/ContinuityAdvElemKernel.C
@@ -105,7 +105,7 @@ ContinuityAdvElemKernel<AlgTraits>::execute(
 
   SharedMemView<double***>& v_dndx = shiftPoisson_ ?
     scratchViews.get_me_views(CURRENT_COORDINATES).dndx_shifted : scratchViews.get_me_views(CURRENT_COORDINATES).dndx;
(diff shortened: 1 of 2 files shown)


sayerhs:
@spdomin There's already this snippet:

```
    // sanity checks; if user asked for shifted Poisson, then user will have reduced sensitivities
    if ( cvfemShiftPoisson_ ) {
      if ( !cvfemReducedSensPoisson_) {
        NaluEnv::self().naluOutputP0() << "Reduced sensitivities CVFEM Poisson will be set" << std::endl;
        cvfemReducedSensPoisson_ = true;
      }
    }
```

This seems to address the setting where shifted Poisson is selected, but reduced sensitivities isn't. Did I misunderstand you?

spdomin:
Oh.. funny. I can approve now. I guess I thought of that user blunder before..

spdomin:
@mbarone81 please update Cori, at your leisure, to reflect this latest bug fix.

PR #146, opened 2017-06-06
Add Gauss-Lobatto quadrature for FEM-based diffusion

spdomin:
* modifed scratch views; since we only have one FEM kernel, it is either
Gaussina quadrature of Gauss Lobatto.

* modified FEM scalar diff to use Gaussian quadrature or GL.

* new GL fem-based verification test

Notes:

a) for efficiency, the FEM kernel probably can be re-written to exploit
GL. However, for now, this is just a study that will evaluate the Laplace
operators on a variety of high aspect ratio meshes.


spdomin:
@rcknaus this one is probably for you. As noted in my notes section, we can later re-write this to avoid some extra interpolations that are not required.

rcknaus:
It looks OK generally.  The ordering of the IPs goes from +1 to -1, though.  Technically there's no issue with that but it's an odd choice that will probably confuse people later on.  We can revisit this later

spdomin:
I wanted this pull request open, however, rejected based on the above comment.

PR #148, opened 2017-06-08
Support all CVFEM diffusion/Laplace operators to be shifted to edge midpoint of element

spdomin:
I would like a review at this point. No need to accept as @rcknaus and I are working from a fork. He plans on adding the boundary master element support.

### DIFF
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -4033,6 +4033,22 @@ Realm::get_noc_usage(
 }
 
 //--------------------------------------------------------------------------
+//-------- get_shifted_grad_op ---------------------------------------------
+//--------------------------------------------------------------------------
+bool
+Realm::get_shifted_grad_op(
+  const std::string dofName )
+{
+  bool factor = solutionOptions_->shiftedGradOpDefault_;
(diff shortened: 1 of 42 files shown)


src/Realm.C
```diff
@@ -4033,6 +4033,22 @@ Realm::get_noc_usage(
 }
 
 //--------------------------------------------------------------------------
+//-------- get_shifted_grad_op ---------------------------------------------
+//--------------------------------------------------------------------------
+bool
+Realm::get_shifted_grad_op(
```
sayerhs: During our `SuppAlg -> Kernel` transition we had discussed removing functionality from Realm and consolidating it in `SolutionOptions`. I think it would be cleaner if we access `get_shifted_grad_op` from `SolutionOptions` directly `realm_.solutionOptions_.get_shifted_grad_op`.

spdomin:
The new kernels access this option from solution options while the old algorithms access from realm. Moreover, the old algorithms access a bunch of stuff in the execute through realm.

In this instance, I was planning on deprecating all of the get_ methods on realm that are on solution options when we deprecate the old algorithms.

PR #149, opened 2017-06-09
Add Gauss-Lobatto quadrature for FEM-based diffusion

spdomin:
* modifed scratch views; since we only have one FEM kernel, it is either
Gaussina quadrature of Gauss Lobatto.

* modified FEM scalar diff to use Gaussian quadrature or GL.

* modify hex8fem isoP location to match internal standard - no mistake,
just more clear.

* new GL fem-based verification test

* fixed a const double npe = nodesPerElement_ in Hex8FEM

* accepted SUPG 1e-15 diff due to IP rearrangement

Notes:

a) for efficiency, the fem kernel probably can be re-written to exploit
GL. However, for now, this is just a study that will evaluate the Laplace
operators on a variety of high aspect ratio meshes.


spdomin:
okay @rcknaus this pull request is good to go subject to your approval.
The quadrature points has been cleaned up in addition to the change that removes the need to specify a specialized kernel creation and, rather, builds on the recent shifted work.

spdomin:
@NaluCFD/core I would like to add this set of operators to our Q3 study. If someone can process this pull request, I would be most appreciative. If anyone has questions, feel free to post them here or through personal email.

PR #151, opened 2017-06-15
Modified PyrSCS node and face mappings to be consistent with changes to

mbarone81:
Face 3 made in recent STK bug fix. Face 3 node ordering changed from
(3,0,4) -> (0,4,3).


spdomin:
[PyramidElementSketches.pdf](https://github.com/NaluCFD/Nalu/files/1079050/PyramidElementSketches.pdf)

spdomin:
okay, let's move on this with the Lagrangian pyramid. If that one still fails, I am open to removing shifted for pyramids... That, or the following:

  // quad4 is the only face that can be safely shifted
  const double *p_intgExp = (face_ordinal < 4 ) ? &intgExpFace_[0] ? &intgExpFaceShift_[0];

      const int row = 9*face_ordinal + k*ndim;
      pyr_derivative(nface, p_intgExp[row], dpsi);

PR #155, opened 2017-06-19
Unit tests for SST, SST/DES and LES Algorithms and SuppAlgs.

marchdf:
As discussed in our standups, here is a set of unit tests for the SST/DES routines. These have been cross-checked with an independent Python script that calculates the same fields/lhs/rhs terms. Please let me know what you think. Thank you!

### DIFF
--- a/unit_tests/algorithms/UnitTestAlgorithmUtils.h
+++ b/unit_tests/algorithms/UnitTestAlgorithmUtils.h
@@ -40,13 +40,16 @@ public:
     const auto& buckets = bulk_.get_buckets(stk::topology::NODE_RANK,
                                             selector);
 
+    double lhs_value = 0.0;
+    double rhs_value = 0.0;
+
     kokkos_thread_team_bucket_loop(buckets, [&](stk::mesh::Entity node) {
-
-      double lhs_value = 0.0;
-      double rhs_value = 0.0;
-
       for (size_t i=0; i < activeSuppAlgs_.size(); ++i){
+	lhs_value = 0.0;
+	rhs_value = 0.0;
+
         activeSuppAlgs_[i]->node_execute(&lhs_value, &rhs_value, node);
+
         lhs_.push_back(lhs_value);
         rhs_.push_back(rhs_value);
       }
--- a/unit_tests/algorithms/UnitTestAlgorithmUtils.h
+++ b/unit_tests/algorithms/UnitTestAlgorithmUtils.h
@@ -53,31 +53,8 @@ public:
       });
   }
 
-  /** Compute norm of the LHS/RHS
-   *
-   * Used to generate the gold values and testing
-   */
-  double norm(const std::vector<double> & vec)
-  {
-    stk::ParallelMachine comm = bulk_.parallel();
-
-    size_t N = vec.size();
-    size_t g_N = 0;
-    double norm = 0.0;
-    double g_norm = 0.0;
-
-    for (int i = 0; i < N; ++i) {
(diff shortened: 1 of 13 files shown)


unit_tests/algorithms/UnitTestAlgorithmUtils.h
```diff
@@ -0,0 +1,92 @@
...
+    const stk::mesh::MetaData& meta = bulk_.mesh_meta_data();
+
+    const stk::mesh::Selector selector = meta.locally_owned_part();
+
+    const auto& buckets = bulk_.get_buckets(stk::topology::NODE_RANK,
+                                            selector);
+
+    kokkos_thread_team_bucket_loop(buckets, [&](stk::mesh::Entity node) {
+
+      double lhs_value = 0.0;
```
sayerhs: @marchdf Should the `lhs_value` and `rhs_value` be re-initialized within the `for` loop below? The SuppAlgs usually do a `+=` operation on the LHS and RHS entries.

marchdf: Good catch. Pushed a fix.

sayerhs:
@spdomin 

The `Realm` and `Simulation` constructors require a `YAML::Node` as an argument. The current setup allows us to tweak the behavior of `Realm::look_ahead_and_creation` call by changing the YAML::Node during `Realm` creation phase. For the data members that are usually set/configured from the `load` method,  the approach is to configure the instance members of `Realm` and `SolutionOptions` in the unit test function directly.

marchdf:
@spdomin I believe @rcknaus has done some unit testing of the projected nodal gradient. I will let him comment on that aspect of the testing. I will work on documenting this for the milestone report and depending on that we can decide what to push into the verification manual.

@rcknaus I am not sure exactly how hard it would be. Everything is pretty tightly integrated for Supplemental Algorithms. This may be the simplest for now. We would probably have to follow something like the Kernel approach?

sayerhs:
@rcknaus Based on our discussions during `SupplementalAlgorithm` -> `Kernel`, that is the long term target (i.e., no dependency on `Realm` for unit testing). 

@marchdf postponed the conversion of Nodal algs to Kernel until we have a prototype for `AssembleSolverAlgorithm` that can loop over all entities, not just elements. 

So I would say this realm thing should be a stopgap measure, not the long term solution.

rcknaus:
@marchdf I just wrote a separate implementation of the projected nodal gradient to use for correctness testing (e.g. that a promoted element can correctly compute the gradient of a linear field in parallel).  We don't have a test for the nodal gradient calculation in Nalu proper.

spdomin:
Agreed. Converting SuppAlgs to remove Realm would look just like Kernels. Basically, providing TimeIntegrator and SolutionOptions. However, we know that essentially all of the nodal algs are deprecated now as we transition to Kernels (as SA notes, node, edge, exposed face and element). I would say that this is a secondary effort and the path forward in this push to create the realm with a YAML input file is good. 

When the team moves forward (next year?) with a full conversion, these unit tests will be re-worked. 

That all being said, my guess is that it would be a one or two day effort to abstract out Realm from all SuppAlgs.

spdomin:
Okay, all of this looks good. Are we ready to commit it then?

marchdf:
@spdomin yep, looks like everyone agrees.

spdomin:
Thanks for handling this. I know how odd it is to process one's own PR:) Therefore, many thanks.

PR #158, opened 2017-06-27
Update header file locations of STK fixtures.

jrood-nrel:
This should fix the errors involving the change of STK header file locations in Trilinos once the fix in https://github.com/trilinos/Trilinos/issues/1433 is merged. Probably doesn't make much difference if we merge this PR now or after the Trilinos fix is merged.

### DIFF
--- a/unit_tests/UnitTestHexElementPromotion.C
+++ b/unit_tests/UnitTestHexElementPromotion.C
@@ -12,8 +12,8 @@
 #include <stk_mesh/base/Field.hpp>
 #include <stk_mesh/base/FieldBLAS.hpp>
 #include <stk_mesh/base/GetEntities.hpp>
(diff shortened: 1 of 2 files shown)


spdomin:
One simple question... If we merge this now, the current builds that are working with an older Trilinos installation will start failing, right? However, if the update of Trilinos occurs, then Nalu will start failing. I would vote that we process this pull request when Nalu starts failing.

jrood-nrel:
Ah, right. I'm so used to using the absolute latest Trilinos. Agreed.

spdomin:
I think you are running the latest Trilinos with the set of patches provided. Once the update of STK is made in Trilinos and provided to develop, your PR should follow.

jrood-nrel:
Trilinos has implemented the fix in the develop branch and it's after the Q3 milestone so I am going to merge this.

PR #159, opened 2017-06-28
WIP. Added NGP Hex8SCS class

spdomin:
I added the HexSCS new calls. I still need to convert HexSCV, however, that should only be a few methods.

At present, I do not have a unit test approach for these new methods. Will work with @sayerhs tomorrow on that.

@alanw0 can use this for unit testing or in crafting the new Solver interface. As noted, anything with SCV_VOLUME will not yet work. That means source terms.

@rcknaus can also massage the newly defined call signatures for grad_op, determinant, etc.

### DIFF
--- a/src/master_element/ngp/hex8/Hex8SCS.C
+++ b/src/master_element/ngp/hex8/Hex8SCS.C
@@ -1,0 +1,144 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+
+#include "master_element/ngp/hex8/Hex8SCS.h"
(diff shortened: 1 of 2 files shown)


src/master_element/ngp/hex8/Hex8SCS.C
```diff
@@ -0,0 +1,144 @@
...
+Hex8SCS::hex8_shape_fcn(
+  const int  &npts,
+  const double *isoParCoord, 
+  SharedMemView<DoubleType**> &shape_fcn)
+{
+  const double one8th = 0.125;
+  const double half = 0.50;
+  for ( int j = 0; j < numIntPoints_; ++j ) {
+    
+    const double s1 = isoParCoord(j,0);
```
alanw0: isoParCoord is passed into hex8_shape_fcn as 'const double*', but appears to be indexed as a 2D view. Is that intended?

spdomin: I must have the wrong version... Sorry. I removed the ngp directory management... Let me make sure I pushed something legit.

PR #160, opened 2017-06-28
WIP Master element HexSCS SIMD

spdomin:
I added the HexSCS new calls. I still need to convert HexSCV, however, that should only be a few methods.

At present, I do not have a unit test approach for these new methods. Will work with @sayerhs tomorrow on that.

@alanw0 can use this for unit testing or in crafting the new Solver interface. As noted, anything with SCV_VOLUME will not yet work. That means source terms.

@rcknaus can also massage the newly defined call signatures for grad_op, determinant, etc.

### DIFF
--- a/src/master_element/MasterElement.C
+++ b/src/master_element/MasterElement.C
@@ -669,14 +669,14 @@ HexSCS::hex8_shape_fcn(
   const double *isoParCoord, 
   SharedMemView<DoubleType**> &shape_fcn)
 {
-  const double half = 0.50;
-  const double one4th = 0.25;
-  const double one8th = 0.125;
+  const DoubleType half = 0.50;
+  const DoubleType one4th = 0.25;
+  const DoubleType one8th = 0.125;
   for ( int j = 0; j < numIntPoints_; ++j ) {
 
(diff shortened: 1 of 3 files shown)


src/master_element/MasterElement.C
```diff
@@ -641,6 +641,317 @@ HexSCS::ipNodeMap(
   return &ipNodeMap_[ordinal*4];
 }
 
+
+//--------------------------------------------------------------------------
+//-------- shape_fcn -------------------------------------------------------
+//--------------------------------------------------------------------------
+void
+HexSCS::shape_fcn(SharedMemView<DoubleType**> &shpfc)
```
sayerhs: @spdomin Any specific reason to maintain this function call instead of implementing the body of `hex8_shape_fcn` here?

alanw0:
I have a commit nearly ready where I demonstrate making the momentum-nso kernel test use AssembleElemSolverAlgorithm rather than going through TestKernelDriver. The reason for this is so that I can make a similar test that uses a new AssembleElemSolverAlgorithmNewME prototype (that uses the new master-elements). But to do that I think I need these changes from @spdomin to be pushed.

PR #161, opened 2017-06-28
Enable OpenMP Threading in Nalu

sayerhs:
- Switch sierra::nalu::DeviceSpace to Kokkos::DefaultExecutionSpace

    - Fix all work arrays in Kernels to be thread safe

    - Update sumInto method to force atomic updates when running in
      threaded mode

### DIFF
--- a/include/BuildTemplates.h
+++ b/include/BuildTemplates.h
@@ -14,15 +14,23 @@
 namespace sierra{
 namespace nalu{
 
-#define INSTANTIATE_KERNEL(ClassName)                             \
+#define INSTANTIATE_KERNEL_3D(ClassName)                          \
 template class ClassName<AlgTraitsHex8>;                          \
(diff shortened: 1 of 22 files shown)


sayerhs:
@alanw0 , @spdomin Wanted to get your opinion on the modifications to enable threading in `AssembleElemSolverAlgorithm`. The code passes regression tests with `OMP_NUM_THREADS=1` and there are a failures `O(6e-15)` differences when running with `OMP_NUM_THREADS=3`. For example,

```
..dgNonConformalElemCylinder........... FAILED:  88.8495  s  max diff:  .00000000000000592999973669620928
```

Curious to hear your thoughts on how we should proceed to test with different threads, and comments on the code.

alanw0:
The diffs that are order machine precision are probably inevitable. Perhaps we need to loosen a testing tolerance a little bit? I'll defer to Stefan on that.

PR #163, opened 2017-07-14
CmakeLists Modifications

jrood-nrel:
I would like to discuss the possibility of allowing Nalu to stand on its own a bit regarding Trilinos. This pull request would add an option to either use the compilers and flags used in Trilinos (by default) or to be able to use any chosen compiler.

It also adds the ability for Nalu to use shared libraries for Yaml.

Mainly I have problems with Nalu forcing the Trilinos configuration because Spack uses compiler wrappers and when I try to build my own deverloper version of Nalu manually on a Cray, it uses the Spack compiler wrappers (on top of the Cray wrappers) that built Trilinos without having the environment set up correctly for them. This can be resolved and possibly other issues people may have in the future by allowing the decoupling of the Trilinos configuration.

I would also add that it allows for a debug build of Nalu without having a debug build of Trilinos.

Note: I have tested this a while ago, but I would like to test it again before any possible merge. I just wanted to submit this for discussion in the meantime.


jrood-nrel:
Closing this. I will need to think more about it if I want to do this right.

PR #164, opened 2017-07-17
Fix reference pressure at an interior node

sayerhs:
Allow user to fix the reference pressure at an arbitrary location within
the computational domain during simulations.

This capability is important for the pressure Poisson solve during
ABL precursor simulations where the side boundaries are usually set to
periodic and the top boundary is specified with a symmetry BC to allow
the flow to adjust naturally when targeting a specified wind
speed/direction at hub height.

### DIFF
--- a/include/Enums.h
+++ b/include/Enums.h
@@ -28,7 +28,17 @@ enum AlgorithmType{
   NON_CONFORMAL = 10,
   ELEM_SOURCE = 11,
   OVERSET = 12,
-  WALL_ABL = 13
+  WALL_ABL = 13,
+
+  /** Set the reference pressure at a node.
+   *
+   *  Used only for continuity equation system. This needs to be the last
+   *  algorithm applied to the linear system because it resets the row and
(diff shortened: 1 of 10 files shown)


sayerhs:
@mchurchf and @mbarone81 : FYI. 

@spdomin This pull request implements the pressure fix logic as well as incorporates the comments from your email:

- Renamed `ABLFixPressure` to `FixPressureAtNode` 
- Introduced a line command in input file to allow user to specify either the STK Node Identifier or a spatial location (uses nearest node search in this case). 
- Moved the algorithm to `solverDirichAlgMap` so that it is executed after all the interior and BC algorithms are executed.

spdomin:
@sayerhs Do you have an example of how this is run somewhere? I can probably back it out from the code. I think that you could try modifying a rtest as well.

sayerhs:
Take a look at line 133 in the attached file:

[ablfixpressure.txt](https://github.com/NaluCFD/Nalu/files/1153628/ablfixpressure.txt)

PR #170, opened 2017-07-20
Activate mdot correction at open bcs

spdomin:
* add global correction to LHS open BC.

* when mdot correction is active:

  1)  no pStab is used. As such, zero out open
      pStab contribution to LHS/RHS in addition to OpenMdot.

  2) make sure pressure_bc is neither registered nor used in the code
     base. This, importantly, includes Gjp contributions as well.

* ensure that at inflow bcs, cont_velocity_bc is neither registered nor used

### DIFF
--- a/include/ComputeMdotAlgorithmDriver.h
+++ b/include/ComputeMdotAlgorithmDriver.h
@@ -28,6 +28,8 @@ public:
   ~ComputeMdotAlgorithmDriver();
 
   double compute_accumulation();
+  void correct_open_mdot(const double finalCorrection);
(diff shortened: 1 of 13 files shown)


mchurchf:
I see no problems in this implementation.  This is a useful approach, but replacing T_ref with <T> in the Boussinesq buoyancy term will probably allow the use of the current open bc.  However, if we use variable density, then we'll need this unless we can come up with an improved local open approach.

spdomin:
This pull request is now out-of-date.

PR #171, opened 2017-07-21
Convert Ksgs source term for usage in SIMD branch

spdomin:


spdomin:
@sayerhs we are having issues rebasing the simd branch.

PR #172, opened 2017-07-25
Hex8 SIMD MasterElements

rcknaus:
All tests pass on machine except for two:

..dgNonConformalElemCylinder........... FAILED:  57.4633  s  max diff:  .0000000000058842900001090502754

..elemHybridFluids........... FAILED:  88.1346  s  max diff:  .00000000000000260400003782159739

Both are ~1.0e-12.  Pull request doesn't include a rebless.


src/MixtureFractionEquationSystem.C
```diff
@@ -387,11 +387,10 @@ MixtureFractionEquationSystem::register_interior_algorithm(
     stk::topology partTopo = part->topology();
     auto& solverAlgMap = solverAlgDriver_->solverAlgorithmMap_;
     
-    AssembleElemSolverAlgorithm* solverAlg = nullptr;
+    AssembleElemSolverAlgorithmNewME* solverAlg = nullptr;
```
spdomin: I think that we need NewME to be the AssembleElemSolverAlg name and deprecate the other...

PR #173, opened 2017-07-26
Hex SIMD MasterElement

rcknaus:
* Three tests failures, all ~1.0e-12.  There's a slight order-of-operations difference here in how the gradient is computed.  The diff in milestoneRunConsolidated is due to [this commit](https://github.com/rcknaus/Nalu/commit/653847b6ca93d026cf78be773eb07cff4a4fa418).

  *   milestoneRunConsolidated: max diff: .00000000000000200002774758401002
  *   dgNonConformalElemCylinder: max diff: .00000000000588429000010905027540
  *    elemHybridFluids:  max diff: .00000000000000260400003782159739

* Small optimization in Ksgs
* Use the "NewME" assemble elem solver fully, with a flag that determines whether the interleaving should be done on the master elements.


spdomin:
great job Robert. I will start testing and give @sayerhs the go ahead on pushing this to master should the performance data look good at scale.

PR #174, opened 2017-07-27
Nalu Docs: Create CMake rule for generating PDF documentation

sayerhs:
- New target `sphinx-pdf` that calls the `latexpdf` sphinx generator
      to create a Nalu.pdf file.

    - Fix RST declarations of some equations that weren't converted to
      latex properly. Disabled automated wrapping and manually created
      the equations in latex format to get the desired layout.

    - Also added a `sphinx-html` target to bypass doxygen runs - useful
      looking a HTML docs while working on RST files.

### DIFF
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -24,6 +24,19 @@ if(SPHINX_FOUND)
         )
     add_dependencies(sphinx doxygen)
     add_custom_target(docs DEPENDS sphinx)
+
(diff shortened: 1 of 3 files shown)


sayerhs:
@spdomin @jrood-nrel This pull request adds CMake rules to generate PDF documentation via `make sphinx-pdf` see attached file for an example

[Nalu.pdf](https://github.com/NaluCFD/Nalu/files/1180312/Nalu.pdf)

PR #175, opened 2017-07-28
Add a Boussinesq buoyancy source term based on a time-averaged ref. temperature

rcknaus:
* Activate with "buoyancy_boussinesq_ra".  Need to specify a "boussinesq_time_scale" value in user constants for this to work. 

* Need to somewhat-carefully choose the time scale for this model to make sense--longer than the turbulence timescales at least.

* Should allow setting a p = 0 boundary condition even with a capping inversion.

* The temperature is filtered using a running average (exponential).  It's computed directly before the momentum solve but after the initial condition output.  So, like the diffusion coefficient or scalar dissipation, the time_filtered_temperature isn't populated at the initial condition output.


rcknaus:
OK.  Added time_filtered_temperature to the output.  "time_filtered_temperature", as I have it, is first populated [right before the first momentum solve](https://github.com/NaluCFD/Nalu/pull/175/files#diff-f5f5cc287590f3a399d33b287d4c4014R617).  So it'll appear as "zero" in the initial condition output, but it'll use the initial temperature distribution in the actual source term calculation on the first iteration.

mchurchf:
Here is a more fundamental question about how to do this:  Really, by doing this, you are saying that what you're subtracting from pressure to get motion pressure is \int_0^z <\rho> g dz.  So, then Boussinesq or not, your gravity term ends up looking like (\rho - <\rho>)g.  For Boussinesq, \rho is constant, but the "\rho" that goes into the gravity term is not constant, so let's call it \rho_b.  In that case, your gravity term is (\rho_b - <\rho_b>)*g.  Boussinesq says that \rho_b = \rho_ref*(1 - \beta (T-T_ref)), and <\rho_b> = \rho_ref*(1-\beta*(<T>-T_ref)).  Plug those two things into (\rho_b - <\rho_b>*g and you get \rho_ref*(-beta*(T - <T>)).

So, not to be a pain in the ass, but wouldn't the more general approach be to do the <>-average of the density that goes into the gravity term?  I ask this looking forward to moving beyond Boussinesq maybe at some point.  If we use just the normal compressible formulation where \rho = p_ref/RT or if we do something anelastic where \rho = p(z)/RT, then we'll end up with the same issue with the open  BC in which we can set p_motion = 0 on the boundary.  However, if we define the motion pressure as I said above and have a gravity term that is (\rho - <\rho>)g, then we can use the standard open bc with p_motion  = 0 on the boundary.

Should we talk on the phone?

spdomin:
Sorry, I am not sure that I like this design. 

The set of code changes adds a new notion of averaging, "moving average" that is its own set of rules outside of the turbulence averaging approach. 

Moreover, it is tied to the momentum equation for execution. The location for this is not even correct, IMHO, as the averaging will take place on an incremental solution, i.e., non-converged.

I thought that the cleanest location for this work would have been within the current TurbulenceAveraging block. If the user asks for this buoyancy option, we can "look ahead" and create a turbulence averaging block on his or her behalf.

So, immediate questions are as follows:

1) why did this not fit well in the TurbulenceAveraging post processing section
2) for correctness, why do we take a moving average on a non-converged solution? It seems like intermediate temperature solutions should not be contributing.

spdomin:
Let's all tune into the stand-up. If you want to start 15 minutes earlier, that will work since we have a parking lot scheduled. I like the more general approach to consider variable density.

sayerhs:
@spdomin I am not sure the logic in [TurbulenceAveraging](https://github.com/NaluCFD/Nalu/blob/master/src/TurbulenceAveragingPostProcessing.C#L436) for the <T> averaging because of the reset logic (line 436 onwards). If we choose a 1 minute or 10 minute averaging window, the average gets reset and is not continuous.

rcknaus:
@spdomin Pushed an update integrating the moving average into TurbulenceAveraging so that it can be used more generally.

spdomin:
This looks much better. Many thanks. Let's wait to see how the first ABL turns out. If it looks good, we can push the code. I am taking the same approach with the global correction scheme so that we do not start admitting research efforts in the code base. We can talk about this strategy at any stand-up.

michaelasprague:
@rcknaus -- would you be able to address the conflicts with Master so that we can merge?

mchurchf:
I approve.  I think Stefan's mdot correction to the open BC is cleaner for flat terrain, but I'm thinking Robert's solution will be useful in complex terrain.

rcknaus:
See https://github.com/NaluCFD/Nalu/pull/284

PR #177, opened 2017-08-24
Hex 27 SIMD master elements

rcknaus:
* Gets promoted elements working with the consolidated approach.  
  - Still need to convert some remaining verification supplemental algorithms (e.g. for the variable density MMS case) to kernels in order to fully get rid of the deprecated ElemSolver methods 
  - Only support polynomial order 2-5 by default for promoted elements.  The user will need to recompile to choose a higher polynomial order
  - Reduced the polynomial order for the kovasznay test from 7 to 5

* Separated out MasterElement to some separate files.  The factory methods at the beginning are now in MasterElementFactory
  - Hex8CVFEM and Hex27CVFEM for hex topologies


spdomin:
What is the state of the regression test suite? If everything passes I
would feel more comfortable.

Design looks good. As a matter of process I might like to shift such design
changes to master so that the merge back looks less like a new code.

rcknaus:
Everything passes on gcc.  On intel, there's still a tolerance issue with the MomentumKernelHex8Mesh.nso_2nd unit test, but this pull request doesn't affect that.  Otherwise, everything passes on both platforms.

spdomin:
Sounds good. We can discuss at the stand-up and then proceed with the pull request. I have some high level design questions best addressed in person. After this, we should clean-up the SIMD branch and sync with master. After that, close the story on Jira.

spdomin:
Okay, I am about to process this pull request. This means that we should start working the merge back. I can try to get a higher order (P=2) milestone R3 mesh done today of tomorrow.

Once we merge back, about two or three stories can close.

PR #178, opened 2017-08-28
Nalu Catalyst integration tests and documentation

tjotaha:
New CMake option ENABLE_PARAVIEW_CATALYST, OFF by default

When enabled, set CMake variable PARAVIEW_CATALYST_INSTALL_PATH
to point to install location of the Trilinos ParaView Catalyst IOSS
Plugin.

Building with Catalyst enabled will create a bash script in
the build directory called naluXCatalyst. This script sets the
environment variable CATALYST_ADAPTER_INSTALL_DIR before starting
naluX.

Modified regression tests ablForcingEdge, mixedTetPipe, and steadyTaylorVortex
to create Catalyst enabled versions. The Catalyst versions end in "Catalyst",
and can be run using ctest -R Catalyst. The pass fail criterion is if the proper
number of output images were created by Catalyst.

Wrote build instruction for ParaView Catalyst and the IOSS Catalyst Plugin in
the Nalu manual Linux build instructions. Added do-config* bash scripts for
assistance in building Catalyst.


tjotaha:
@spdomin and @mbarone81. Could you guys take a look at this pull request? I've added some documentation, tests, and script for Catalyst.

spdomin:
@sayerhs and @jrood-nrel, could you please comment on this pull request as it may require an update on your end? This looks like a new build dependency which I need your opinion.

jrood-nrel:
I really appreciate that this PR was thorough and has documentation. I have a few comments though:

The regression tests are not lightweight tests. Do we feel it is necessary to duplicate 3 of the regression tests into additional Catalyst tests? Can we just merge the Catalyst tests with existing tests and have CMake run the 3 tests with Catalyst if it's turned on, or just run them as normal if not? Or maybe just 1 additional Catalyst test would suffice? Also if a user doesn't have Catalyst, it would be nice if they knew the Catalyst test(s) were skipped, and I think CTest has a way to skip tests. I'm also not certain the added tests qualify as regression tests unless we are at least checking some kind of metrics on the images Catalyst is outputting. Are these tests meant more to be examples of how to use Catalyst? Should that just go into the documentation?

Another thing I'm wondering is if there is an official `FindCatalyst.cmake` that can be used?

Lastly, will we need to track the latest Catalyst to test it effectively?

In the end, the extra dependency is fine with me if it is necessary for a great feature like in situ visualization (assuming Catalyst doesn't need to be updated too often), but I'm mostly worried about the large additional testing time here because resources are an issue and the burden increases in multiple dimensions.

spdomin:
In the stand-up today, a question regarding coordination of Catalysts with the NREL team lead (Kenny) was brought up. We may want to take this discussion thread offline to define the best path forward.

tjotaha:
Yes, I agree that some coordination would be helpful. Jeff Mauldin has been corresponding with Kenny regarding getting a Catalyst enabled Nalu build up and running at his site.

tjotaha:
We can locate ParaView and check for the Catalyst Python dependencies with a CMake line like the following:

find_package(ParaView REQUIRED vtkPVPythonCatalyst)

I wrote the build instructions assuming that the target system has nothing installed. It makes sense to check for ParaView in the Nalu CMake. If ParaView is found and Nalu use Catalyst is enabled, then build the Nalu Catalyst IOSS plugin from Trilinos and install it.  The fallback would be to have the user build everything manually if ParaView is not found and Nalu Catalyst is desired.

The Nalu Catalyst integration does depend on specific versions of ParaView, and will need to be tracked for testing purposes.

tjotaha:
The 3 Catalyst regression tests are only created when ParaView Catalyst is enabled in Nalu. For Catalyst with Sierra, we do run image difference tests to make the tests true regression tests. The image difference requires an additional dependency during the build that I wanted to avoid. The Catalyst regressions tests serve to give some confidence that Nalu Caltalyst is installed and working properly, and they can be used as examples of how the integration works. We are working on creating additional documentation for the Catalyst interface that defines the visualization.

jrood-nrel:
Ok, I will let you coordinate with Kenny. I think the only thing I would really like to see in this before it's merged would be to add Catalyst into a few existing tests, rather than add 3 more tests for it. Other than that, I don't have a strong opinion on merging this.

spdomin:
The code repos is broken. If we can wait on the merge, that would be great.

tjotaha:
Okay. I'm working on moving the Catalyst regression tests into the existing tests.

tjotaha:
I moved the 3 Catalyst regressions tests into their corresponding existing regression tests and updated the documentation.

jrood-nrel:
Thanks @tjotaha .

tjotaha:
@sayerhs, @jrood-nrel, or @spdomin were one of you guys going to merge this branch? It looks like I don't have permission to do it.

jrood-nrel:
I will work to get the catalyst tests turned on an integrated into our testing. Thanks @tjotaha

jrood-nrel:
@tjotaha After merging this, it now looks like many of the tests are silently broken. They show that they are passing, but they aren't running many of the tests. Could you help me fix this? Can we have a separate test category for the 3 tests that have catalyst capability and let the remaining tests use the original regression add_test function?

PR #179, opened 2017-08-28
Use custom sumInto function instead of Tpetra call

rcknaus:
* Uses a sorted list of local ids to simplify avoid some work the Tpetra suminto performs

### DIFF
--- a/include/LinearSystem.h
+++ b/include/LinearSystem.h
@@ -70,8 +70,11 @@ public:
       const SharedMemView<const double*> & rhs,
       const SharedMemView<const double**> & lhs,
       const SharedMemView<int*> & localIds,
+      const SharedMemView<int*> & sortPermutation,
       const char * trace_tag
       )=0;
+
+
 
   virtual void sumInto(
     const std::vector<stk::mesh::Entity> & sym_meshobj,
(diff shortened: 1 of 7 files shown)


spdomin:
Okay, is this ready to go or do we need to see a formal response from @rcknaus?

rcknaus:
For performance, I saw about a 15% speedup on milestonRunConsolidated and about 10% speedup on hoHelium both run locally.  From vtune, it looked like a 3x speedup for sum_into_row vs sumIntoValues on milestonRunConsolidated.

I'm not really sure what to do about #3.

alanw0:
@rcknaus those sound like great performance improvements.
You don't need to do anything about #3, I'll close the Jira story.

sayerhs:
@rcknaus Can you resubmit this against the `master` branch? Since the SIMD pull request was closed, I would like to avoid further commits on top of that.

spdomin:
We also need to have a performance sanity check, ideally, on a moving mesh simulation. My V27 mesh would be ideal. I want to make sure that setup costs do not change under this new formulation. Also, I was thinking that the change would only require sumInto to be modified, however, I see new data structures here and there. This is something I would like to understand better.

PR #180, opened 2017-08-28
Use local kokkos matrix API instead of Tpetra::CrsMatrix API for

alanw0:
replacing values.

We use replaceValues for applying dirichlet boundary conditions.
We have been given guidance from the Trilinos folks to prefer using
the Kokkos local matrix rather than the Tpetra::CrsMatrix for
modifying matrix values.

### DIFF
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -1215,6 +1215,7 @@ TpetraLinearSystem::applyDirichletBCs(
   stk::mesh::BucketVector const& buckets =
     realm_.get_buckets( stk::topology::NODE_RANK, selector );
 
+  const bool internalMatrixIsSorted = true;
   int nbc=0;
   for(const stk::mesh::Bucket* bptr : buckets) {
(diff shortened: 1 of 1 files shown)


sayerhs:
@alanw0 Should this pull request also replace lines [1312](https://github.com/alanw0/Nalu/blob/739d0f59ee8ccf03d8eacf50c5735c050f73afb1/src/TpetraLinearSystem.C#L1312) and [1357](https://github.com/alanw0/Nalu/blob/739d0f59ee8ccf03d8eacf50c5735c050f73afb1/src/TpetraLinearSystem.C#L1357). What about the corresponding `replaceLocalValues` in the `RHS` vector?

alanw0:
@sayerhs thanks, I will add another commit to replace those two usages of replaceLocalValues that I missed. As for the vector occurrences, that's not as clear. I'll check whether there's an equivalent substitution we should do for vectors.

alanw0:
ok, I've added another commit which changes the remaining calls of replaceLocalValues.

PR #182, opened 2017-08-30
Changed the behavior of residual norm reporting to accomodate the use

mbarone81:
case where the user specifies max_iterations > 1 for individual
equation systems.  This fixes a bug that caused an inappropriate
scaling of the nonlinear residual by the current nonlinear residual
only if the outer real iteration counter was at 1.  A new equation
system iteration counter has been introduced, and now the nonlinear
convergence behavior is tracked and reported in the EquationSystem
class rather than the TpetraLinearSystem class.

Added a sysName_ variable to the EquationSystem class.  This was done
to preserve residual norm output format while transferring control of
the output to the EquationSystem class.

Preserve special-case residual output for PNG, RTE, PMR equation sets, using the
reportMyResiduals_ variable within the EquationSystem class.

### DIFF
--- a/include/EquationSystem.h
+++ b/include/EquationSystem.h
@@ -224,12 +224,15 @@ public:
   void report_built_supp_alg_names();
   bool supp_alg_is_requested(std::string name);
   bool supp_alg_is_requested(std::vector<std::string>);
-
+  
+  void reset_nonlinear_iteration_count();
   bool nodal_src_is_requested();
 
   EquationSystems &equationSystems_;
   Realm &realm_;
   std::string name_;
+  std::string sysName_;
+  bool reportMyResiduals_;
   const std::string eqnTypeName_;
   int maxIterations_;
   double convergenceTolerance_;
@@ -249,6 +252,9 @@ public:
   int nonLinearIterationCount_;
   bool reportLinearIterations_;
   bool edgeNodalGradient_;
+  int eqSysNonLinearIterationCount_;
+  double firstNonLinearResidual_;
+  double scaledNonLinearResidual_;
 
   void update_iteration_statistics(
     const int & iters);
--- a/src/EquationSystem.C
+++ b/src/EquationSystem.C
@@ -58,8 +58,13 @@ EquationSystem::EquationSystem(
     nonLinearIterationCount_(0),
     reportLinearIterations_(false),
     edgeNodalGradient_(realm_.realmUsesEdges_),
+    eqSysNonLinearIterationCount_(0),
+    firstNonLinearResidual_(0.0),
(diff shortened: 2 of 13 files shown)


sthomas61:
Matt - the linear residuals are output instead of the relative residuals - could this be changed because the GMRES solver uses a relative residual tolerance to test for convergence - typically 1e-5. So the residual at convergence is impossible to compare with the relative tolerance

src/EquationSystem.C
```diff
@@ -294,6 +299,32 @@ EquationSystem::assemble_and_solve(
   update_iteration_statistics(
     linsys_->linearSolveIterations());
   
+  // update equation system nonlinear iteration count
+  if (linsys_->linearSolveIterations() != 0)
```
sayerhs: I don't understand why we need an `if` check here. Shouldn't this counter be always updated if we have called the `linsys_->solve` method ?

include/EquationSystem.h
```diff
@@ -224,12 +224,15 @@ class EquationSystem
...
-
+  
+  void reset_nonlinear_iteration_count();
   bool nodal_src_is_requested();
 
   EquationSystems &equationSystems_;
   Realm &realm_;
   std::string name_;
+  std::string sysName_;
+  bool reportMyResiduals_;
```
sayerhs: Can you explain why using `linsys_->provideOutput_` within EquationSystem methods was not possible? Should this variable be removed from the code now that we have `reportMyResiduals_`?

spdomin:
Lots of discussion here:)

First, the desire here is to provide a normalized nonlinear residual and use that for outer Picard loop termination. 

Second, the issue here is the following original code:
if ( realm_.currentNonlinearIteration_ == 1)
  // set initial residual value for later usage in norms scaling

currentNonlinearIterations and even eqSysNonLinearIterationCount_ is not specific enough.

The issue is, and still remains, that we desire to save off the nonlinear residual the **first** time it is computed within a time step. Although the code in this pull request will work for the current use case, it will not work for, e.g., iterations over Realms that might occur in CHT, PMR/Fluids, CHT/PMR/FLuids, and, germane to this project, FSI.

I would move forward by adding a bool on the eqs_ that speaks to a first time a non-residual norm is computed. I would reset this bool to false not at the eqVec iteration, but rather, as a pre-time step utility-like interface which is outside the scope of the current modification. Note that a virtual method off of the EquationSystem class also does the trick - we just need it at the outer scope that includes Realm iterations.

If we follow the above procedure, this means that we will allow for the most general use case, specifically, simulations that inlcude multiple realms, multiple non-linear iterations and, finally, multiple equation system iterations within the MyEqs::solve_and_update().

PR #188, opened 2017-09-05
Use custom sumInto function instead of Tpetra call

rcknaus:
* Sorts the local column indices on a per-element basis.  This allows us to search for offsets and modify matrix values in just one pass for a particular element being summed into the linear system.  

* Shows a good speed-up on small tests problems, especially with intel compilers.  I plan to---but have yet to---get performance data on a large scale problem.


sayerhs:
See also discussion in #179

spdomin:
Okay, processed this. When the V27 case is complete, please report back. In the meantime, we may need this functionality for better SIMD/threaded performance.

PR #192, opened 2017-09-07
Change l-o-s from sequence of variables to single row

spdomin:
* modified line-of-sight to output the following:

Time	Coords[0]	myFieldOne[0]	myFieldTwo[0]

* removed mean and variance as per team's suggestion.

### DIFF
--- a/include/DataProbePostProcessing.h
+++ b/include/DataProbePostProcessing.h
@@ -109,8 +109,8 @@ public:
   // populate nodal field and output norms (if appropriate)
   void execute();
 
-  // output the average value
-  void provide_average(const double currentTime, const int timeStepCount);
+  // output to a file
+  void provide_output(const double currentTime);
   
   // provide the inactive selector
   stk::mesh::Selector &get_inactive_selector();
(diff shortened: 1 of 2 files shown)


marchdf:
@spdomin Do you think there is a way to not have it print out a blank line and the banner every time `provide_output` is called (have it only print out the banner once, when the file is first created)? Currently, in the [attached file](https://github.com/NaluCFD/Nalu/files/1285163/probe_profile1_0.txt), a blank line and the banner is reprinted at regular intervals. That information is redundant. A simple if condition around the banner output on whether the file has already been written to should do the trick I think.

marchdf:
@spdomin Blank lines between outputs might actually still be useful for those who use gnuplot (and blank lines are ignored by standard libraries). But the repeated headers do cause unnecessary parser code.

spdomin:
@marchdf, I would be happy to change the banner output. Let me keep the blank lines between the time output and remove the banner. I can make that small tweak and then resubmit or push directly.

One complexity might be on restart since the file is augmented, however, I am sure that I can deal with this use case easily.

marchdf:
Sounds good. Feel free to push directly if you manage to get rid of the repeated banners.

PR #193, opened 2017-09-07
Re-commit "Use local kokkos matrix API instead of Tpetra::CrsMatrix A…

alanw0:
…PI for..."

commit 9437e0c988dd7ec7a3ea3d0ab603e79a06fa9fa5 was previously reverted.

This puts it back in, but also fixes the issue that was seg-faulting
when a zero rowLength was encountered. (Which manifested as indices->getRawPtr()==0)

### DIFF
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -1201,7 +1201,7 @@ TpetraLinearSystem::sumInto(
   )
 {
   const size_t n_obj = entities.size();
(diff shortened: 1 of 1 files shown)


alanw0:
@spdomin I was able to reproduce the throw using the ablUnstableEdge problem as you described. This pull-request fixes it and puts back in the commit that was reverted.

alanw0:
I added a second commit which is minor, it just fixes a couple of compiler warnings.

PR #195, opened 2017-09-07
Change l-o-s from sequence of variables to single row

spdomin:
* modified line-of-sight to output the following:

Time	Coords[0]	myFieldOne[0]	myFieldTwo[0]

* removed mean and variance as per team's suggestion.

* provide banner only once per file, not per time step


spdomin:
Sorry, something off happened here and I needed to re-process the pull request.

PR #198, opened 2017-09-22
Trying to get the cmake code more organized.

jrood-nrel:
Since we are adding more TPLs, I took a little time to try to get the cmake code a bit more organized and consistent for the future.

This _should_ be just a cosmetic thing unless I screwed something up of course, and I tried not to disturb any fundamental way things happen in the Cmakelists, **EXCEPT** I changed how Nalu searches for the YAML library to allow it to be more flexible and find YAML as a shared library as well.

I have tested that this works the same as it did before with how I'm building Nalu at NREL, but it would be great if at least one other person could double check that I didn't mess up something obvious.


reg_tests/CTestList.cmake
```diff
@@ -147,10 +145,12 @@ add_test_r_np(periodic3dEdge 8)
 add_test_r(quad9HC 2)
 add_test_r(steadyTaylorVortex 4 6)
 add_test_r_rst2(steadyTaylorVortex_P4 8)
-add_test_r_rst2(hoVortex_P2 8)
```
sayerhs: This test is being removed as part of this pull request. Is this intentional?

reg_tests/CTestList.cmake
```diff
@@ -147,10 +145,12 @@ add_test_r_np(periodic3dEdge 8)
...
 add_test_r(variableDensNonIso 2)
 add_test_r(variableDensNonUniform 2)
 add_test_r(variableDensNonUniform_P5 8)
+if(ENABLE_OPENFAST)
```
sayerhs: **Wishlist**  It would be nice if the CTest output (as well as CDash) showed information about _skipped tests_ instead of just removing those tests.

reg_tests/CTestList.cmake
jrood-nrel: Relocated.

reg_tests/CTestList.cmake
jrood-nrel: I was going to add this into this pull request but I could only find the following because documentation for CMake is hard to find: https://cmake.org/Wiki/CMake_Testing_With_CTest#Customizing_CTest and I didn't spend the time to test an implementation. I guess I just add the CTestCustom file, add all tests in CTestList, and then use `if(NOT ENABLE_OPENFAST)` to put the test in the `CTEST_CUSTOM_TESTS_IGNORE` list?

michaelasprague: @jrood-nrel @sayerhs  -- is this ready for merger, or is more work required?

jrood-nrel:
I've tested this and I would like to merge it. If it does something for somebody I didn't expect, it should be a quick fix.

sayerhs:
I don't see any issues with merging this.

PR #199, opened 2017-09-22
Remove old nightly build stuff?

jrood-nrel:
We can discuss this at the standup, but I feel like we should remove this stuff. The style guide is in the documentation, and one should look to the CTestNightlyScript if wanting to schedule testing.


spdomin:
As am I per our stand-up discussion with @jrood-nrel

PR #202, opened 2017-09-28
Fix for ParaView Catalyst regression tests

tjotaha:
Added Catalyst script line for CMake configure.


tjotaha:
@jrood-nrel thanks for fixing my last merge request. I ran the Catalyst tests, and noticed there was a missing line to reference the Catalyst script in the Nalu input deck when it is configured by CMake.

jrood-nrel:
@tjotaha Is `CATALYST_FILE_INPUT_DECK_COMMAND` actually necessary? I didn't see it referenced anywhere after it was set in the initial commits for Catalyst so I removed it.

tjotaha:
@jrood-nrel The CATALYST_FILE_INPUT_DECK_COMMAND variable gets set to a string when CMake runs. This string is then substituted in the *.i.in file on the next line during the configure_file(). This produces a Nalu input deck *.i file that will invoke Catalyst when run instead of the normal Nalu Exodus 2 output.

jrood-nrel:
Ah yes. Ok, great. Thanks! I will merge it.

PR #204, opened 2017-09-29
Kokkosized TetSCS::determinant and TetSCV::determinant

alanw0:
Converted these functions to use Kokkos views. This involved
converting some fortran code to C++, etc.

Used the unit-testing infrastructure Robert had started in
UnitTestKokkosME. In this case I'm using it to compare the
new determinant results to the old determinant results.


spdomin:
@alanw0, nice work. I will review this later in the day (or by the latest, tomorrow).. thanks.

rcknaus:
This looks good to me.

One note is that for tets (and tris) the gradient is constant over the element.  So, one can just compute the gradient at the first IP location and then copy the values to the other IPs.  It's probably not a big cost saving, though.

Otherwise, I didn't realize this initially, but some of the geometric calculations for the tet and hex elements are shared.  I think [this function](https://github.com/NaluCFD/Nalu/blob/master/include/master_element/Hex8GeometryFunctions.h#L31) should work for calculating the area vector for the subcontrol volumes of the tet as well.

alanw0:
I'm currently trying to finish the TetSCS::gij method, and I've got something wrong that I'm chasing down with valgrind etc. If you want, let me finish that before merging any of this. valgrind seems to be saying I've got something wrong in grad_op also...

alanw0:
@spdomin and @rcknaus I've now added a commit for the TetSCS::gij method, which I think is the last method needed for the enums we have defined for ElemDataRequests/ScratchViews. I haven't looked into the 'meFEM' enums, perhaps there is more to do there.

spdomin:
The meFEM is for a hex8 FEM-based element. No need to go there. Did you add the plumbing for the creation of the ElemSolverAlg? I am about to look now. I can convert a few tests over to use the consolidated ret-based approach...

alanw0:
@spdomin I did not add the plumbing for the solverAlg. I'll do that next.

spdomin:
Take a look at @rcknaus comment on re-use of some hex triangulation methods. The re-use of certain functions was my first comment.

Second, can you pull all of the Tet methods from MasterElem.C to Tet4CVFEM.C? Add both the SCS and SCV classes there (both new and old). In time, we want MasterElement.C to be broken down to each topo. I suppose any free function can live in MasterElement.C

Finally, when you have the plumbing done, we can merge and move forward with transitioning of tests.

alanw0:
@spdomin ok, I'll make it so.

spdomin:
dgNonConformal3dFluidsHexTet and elemPipeCHT can be a great conversion suite of tests that I can handle. The question is whether or not we want new ones of the "consolidated" flavor or should be not look backward past the first benchmark that I will perform? @sayerhs, what do you think?

sayerhs:
@spdomin Removing old tests should happen in conjunction with removing/deprecating old flavor. If the rest of the team is going to use the code and may be add modifications to it, I would be hesitant to remove those tests. While we test the two cases you mention as part of this branch, but only convert one to consolidated flavor for m erging with master, so that we have exposure to both algorithms in the test suite?

spdomin:
Okay, we can keep the old non-consolidated in favor of a new parallel consolidated flavor. When we finally deprecate the "old" way, we can remove those tests.

In the meantime, my proposed elemPipeCHT includes the  hybrid_factor: velocity: 1.0. As such, I am going to convert the full MUSCL-based advection operator (with diffusion) to the equivalent upw_advection_diffusion that we have for scalars. This should be done by COB or early tomorrow morning. Then, @rcknaus can add his parallel enthalpy test for consolidated cleanly.

alanw0:
@spdomin and @rcknaus I added a couple more commits which address your comments and requested changes. You can either merge like this, or I can squash it all down into one commit and put it back in as a new pull request. Either way is fine, each of these commits is good in the sense that they are each buildable/runnable.

PR #205, opened 2017-10-02
Add consolidated kernels to the EnthalpyEqSys

rcknaus:
* Allows the "use_consolidated" approach for enthalpy as well.


spdomin:
This looks good. However, I would like to see a test converted to accompany this commit. Could you pick a couple element-based non-isothermal cases to convert? Make sure that the errors are low between the former gold and newly converted scheme. Experience suggest machine tolerance is expected.

rcknaus:
Added a new test "nonIsoElemOpenJetConsolidated".  New test passes with the non-consolidated nonIsoElemOpenJet.norm.gold file (but this includes a rebless nonetheless).

spdomin:
I will approve this - well done.

@sayerhs, I wonder about the path forward to create new tests. I still agree that we should, however, wondering if we should be changing the original test and adding myTestDeprecated? The path forward now will note the proliferation of myTestConsolidated. I guess we can move things around later? Not a big deal, just wonder if there is a better protocol to follow in the future.

sayerhs:
@spdomin I agree that changing the original test to use consolidated algorithms and moving the old ones into `TestDeprecated` would be the ideal approach. I will have to loook at the current CTest setup to see if the `add_test_*` can support directory structure (i.e., adopting a structure like `deprecated/TEST`), my guess is no.

PR #208, opened 2017-10-05
Converted Pyramid master-elements to use Kokkos views.

alanw0:
Was able to re-use a couple of functions that the Tet uses,
and also the hex_volume_grandy function.


sayerhs:
@alanw0 I will need the `tet_deriv` and `threeD_gij` for the Wedge6 refactoring also. Should those be part of the utility functions header? 

@spdomin I will wait for this pull request to be merged so that I don't duplicate the functions Alan has for implementing `grad_op` and `gij`.

alanw0:
@sayerhs do you mean the header MasterElementUtils.h ? I can certainly move those functions into there. I think that makes sense, I'll do that and add another commit.

alanw0:
@sayerhs I've moved those functions to MasterElementUtils.h. I renamed tet_grad_op to general_grad_op.

rcknaus:
There is a [generic_grad_op](https://github.com/NaluCFD/Nalu/blob/master/include/master_element/MasterElementFunctions.h#L32) and gij already that you can use if there's no point in specializing the gradient computation.  See the [Hex8CVFEM grad_op](https://github.com/NaluCFD/Nalu/blob/master/src/master_element/Hex8CVFEM.C#L496) for an example.

spdomin:
Agreed. Robert hs a bunch of general methods coded. Should we push all of the current code and then move Roberts general code to the Util header? Let me know as I can move either way.

alanw0:
I can do that this morning. I'll add another commit to the pull request that does that.

alanw0:
ok @spdomin @rcknaus @sayerhs I think it's ready to go now. I added a commit which removes the redundant grad_op and gij functions I had translated from fortran, and instead uses the generic ones that Robert had already put in MasterElementFunctions.h.

PR #210, opened 2017-10-09
Converted Quad4 master-elements to use Kokkos views.

overfelt:
My first Nalu pull request, still learning how to do this on GitHub.

This is part of the Kokkos conversion for master elements and just mimics what Alan Williams did for tetrahedral elements and applies it to quadrilateral 4 elements.
(1) The unit test was extended to cover this new class
(2) New class is implemented in new header and source files
(3) Old class is deleted
(4) All regression tests (73 of them) were run optimized and pass


spdomin:
@overfelt, we have a bunch of pull requests for master element re-factoring. After Shreyas' pull request on Wedges is accepted (he is fixing merge conflicts now), go ahead and re-submit.

alanw0:
@overfelt James I'm off on the stk project for a sprint now so I won't be as plugged into this unfortunately. But Robert Knaus @rcknaus is a good reviewer for master-element work, he knows them inside and out.

overfelt:
Stefan,

So I should copy the UnitTestHexMasterElement test to create one for Quad?

So far, I just added the new quad 4 to the UnitTestKokkosME.C test which just compares the old fortran versions of a couple of functions to the results from the new C++ kokkos enabled versions to see they are the same.  There is also a simple is-in-element test for quads that uses the new class. 

jro

include/master_element/Quad42DCVFEM.h
```diff
@@ -0,0 +1,194 @@
...
+// 2D Quad 4 subcontrol volume
+class Quad42DSCV : public MasterElement
+{
+public:
+  Quad42DSCV();
+  virtual ~Quad42DSCV();
+
+  const int * ipNodeMap(int ordinal = 0);
+
+  void determinant(
```
sayerhs: @overfelt With the merge of Alan's Pyramid NGP modifications, the method signature has changed to pass-by-reference instead of pass-by-value.

spdomin:
@overfelt, well done:) I think that you are clear for the tri3 elements. This should be the last one and you should be free from any merge conflicts.

PR #213, opened 2017-10-11
Tri32 dcvfem

overfelt:
Continuing the task of adding master element functions that take Kokkos views.  This is the three node 2D tri.


spdomin:
Up for the quad9:)

spdomin:
@overfelt, @sayerhs, @rcknaus and @alanw0 - I am starting to convert a few tests and finding some issues. The one I am on is the elemHybridShift rates. If I add consolidation and activate upw_adv_diff for mixture fraction, the case throws... Fig I cane back to adv_diff for mixture fraction, it runs.. Any ideas? I am sure I can figure this out, just wondering at a high level.

alanw0:
No specific ideas... sounds like a task for the debugger, see where and why it's throwing. If it's throwing due to some numerical issue that accumulates, that would be tricky. The smaller the test the better. Can you try it on a toy mesh?

spdomin:
All I have thus far is:

terminate called after throwing an instance of 'std::runtime_error'
  what():  View bounds error of view  ( 0 < 0 , 0 < 0 , 0 < 0 )

in a view bounds verification...

alanw0:
@spdomin It could be that a view is being accessed but the pre-req wasn't declared. If you can set a breakpoint on '__cxa_throw' in totalview it should stop where the view is being accessed and that throw is occurring.

spdomin:
What is this?

bytes_per_thread *= 3*simdLen;//3 is wrong, still need to debug this!!! should be 2 but seems to have memory problems.

alanw0:
the bytes_per_thread thing was something that should never have been pushed like that, I thought I had already taken care of that... It means that 3*simdLen is an over-allocation, reserving more memory than should be necessary. A sharper bound was thought to be 2, but some cases didn't run successfully with that. So there is still some counting inconsistency in there related to how the views are getting inflated.

spdomin:
Alan and I found the issue... Will close down this discussion here...

Issue #220, opened 2017-10-17
VLA array used in Tri32DCVFEM.C


sayerhs:
```
Nalu/src/master_element/Tri32DCVFEM.C:465:19: warning:
      variable length arrays are a C99 feature [-Wvla-extension]
  DoubleType det_j[nint];
                  ^
1 warning generated.
```

overfelt:
Shreyas, 

I pushed a one line fix for this warning but did not get an email about the push even though my project is now up to date with master.  Did the push fail or am I just not receiving emails?

sayerhs:
@overfelt It appears that you pushed it to your [local fork](https://github.com/overfelt/Nalu/commit/54ca538853c1886f2ad650fc5248c8525a9e3f6f) and not to NaluCFD/Nalu repo.

overfelt:
Drat.

PR #221, opened 2017-10-17
Fix variable lenght array to use std::vector.  #220

overfelt:
#220 Variable length arrays are non-standard even
though some compilers do not flag them as such.

### DIFF
--- a/src/master_element/Tri32DCVFEM.C
+++ b/src/master_element/Tri32DCVFEM.C
@@ -462,7 +462,7 @@ void tri_gradient_operator(
   const int nint = deriv.dimension(0);
   const int npe  = deriv.dimension(1);
  
-  DoubleType det_j[nint];
+  std::vector<DoubleType> det_j(nint); 
 
   DoubleType dx_ds1, dx_ds2;
   DoubleType dy_ds1, dy_ds2;


sayerhs:
@overfelt The fix looks good. However, is it possible to make this a stack allocation instead of initializing the `std::vector` everytime this function is called? Might require this function to be a method of the class so that you can use `numIntPoints_` or templated on AlgTraits?

rcknaus:
I would just change it to a scalar value of detj instead of an array or std::vector.  We don't use detj outside of the scope of the integration point loop (and detj for a triangle element is a constant).

overfelt:
@sayerhs I tried numIntPoints_ but it is not a constexpr so is still a problem.  But since numIntPoints_ is set to 3 in the constructor, we could just hardcode all array that use it to 3?  

I see that some of the master elements use AlgTraits.  Are numScsIp_ and numScvIp_ replacements for the numIntPoints_ defined on the class?  Should I be using these everywhere?  I see there is also nDim_ and NodesPerElement_ defined in both the AlgTraits and as member data.  It seems we could be using the constexpr values in all derived class code.

sayerhs:
@overfelt @rcknaus 's suggestion is the better option instead of using arrays.

overfelt:
Yes, should have been a scalar all along.  Now if I can figure out how to update an open pull request with the new code...

alanw0:
If a pull request is set to pull a change from your fork to the NaluCFD/Nalu repo, then you can update that by simply adding another commit to your fork.

PR #226, opened 2017-10-23
Memory bloat in bloating algorithm bug fix

spdomin:
* Make sure to clear the opposing vector of possible candidate
faces for each dgInfo object.

* see: https://jira.exascaleproject.org/browse/ADSE07-248


spdomin:
Reviewers, please feel free to squash and merge should the changes look good.

spdomin:
@alanw0, thanks for the comments. Let's take another pass at this algorithm under the Jira task to increase the matrix stencil to avoid preconditioner setup costs. At present, the vector does not grow, which is what was killing the runs before.
 
You can also provide a pull request to ensure that the vector is as small as needed. In general, these opposing faces are O(5) in size.

PR #233, opened 2017-11-01
Initial simple unit-tests of the getDofStatus function.

alanw0:
Made the body of getDofStatus into a free-function that is called
by the TpetraLinearSystem class member. That way it can also be
called by unit-tests.

These first unit-tests are very basic, only checking owned and
shared. It will be useful to add slightly more sophisticated tests
to nail down getDofStatus behavior for ghosted, periodic and
other dof flavors.

### DIFF
--- a/unit_tests/UnitTestGetDofStatus.C
+++ b/unit_tests/UnitTestGetDofStatus.C
@@ -1,0 +1,47 @@
+#include <gtest/gtest.h>
+
+#include <UnitTestHelperObjects.h>
+#include <stk_mesh/base/BulkData.hpp>
+
+#include "UnitTestUtils.h"
+
+#include <TpetraLinearSystem.h>
+
+TEST_F(Hex8MeshWithNSOFields, getDofStatus_basic)
+{
+  fill_mesh_and_initialize_test_fields("generated:1x1x2");
+
+  unit_test_utils::HelperObjects helperObjs(bulk, stk::topology::HEX_8, 1, partVec[0]);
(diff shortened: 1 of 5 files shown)


unit_tests/UnitTestGetDofStatus.C
```diff
@@ -0,0 +1,47 @@
+#include <gtest/gtest.h>
+
+#include <UnitTestHelperObjects.h>
+#include <stk_mesh/base/BulkData.hpp>
+
+#include "UnitTestUtils.h"
+
+#include <TpetraLinearSystem.h>
+
+TEST_F(Hex8MeshWithNSOFields, getDofStatus_basic)
```
sayerhs: Any specific reason to use this with `Hex8MeshWithNSOFields` instead of a custom class? I would prefer it to be a separate class that can be used for filtering with `--gtest-filter`

alanw0: That's good feedback, I'll make it a separate class.

alanw0:
@sayerhs I just added a commit which makes the unit-tests use a different class for setting up the mesh. This is a good change. If we pursue this further we'll add code to do ghosting etc. to do stronger testing of getDofStatus.

sayerhs:
@alanw0 This looks good to me. Definitely agree we should create different shared/ghosted entities to test all of DofStatus.

spdomin:
Is this ExaWind work or A2e?

PR #234, opened 2017-11-04
Stefan this is just a conversation starter. (Don't pull/merge.)

alanw0:
Let's chat about this. This commit contains the optional bool for
apply_coeff to ignore the incoming lhs. That's the easy part.

It also contains a possible approach for setting a bool on algorithms.
It is implemented in just a couple of algorithms, and demonstrated in
the unit test UnitTestAssembleOnlyRhs.C.

As I said in email, another approach is to add an optional bool to
SolverAlgorithm::execute, but I didn't end up thinking that made sense
for the Algorithm base class (after I had snaked it through about 30
SolverAlgorithm derivations). Again, let's chat about this.


alanw0:
Stefan, using my new git fu, I think I was able to replace the previous 2 commits with a single commit which now also implements the things we discussed on the phone and which you put in your notes. Take a look and let me know what you think.

alanw0:
@spdomin I've now added the plumbing for applyDirichletBCs and prepareConstraints, and made sure that the rest of the SolverAlgorithm derived classes now pass the bool through correctly to apply_coeff etc.

spdomin:
The code looks good. I think we need to capture the use case for Muelu setup. I just made this comment in Jira; as such, we are crossing tools:)

Also, before we push this code, I want to make sure that algorithmically, it works and is something that we want to pursue. If you like to start a development branch, that would work. Alternatively, I am happy to work off of your fork to evaluate the changes.

alanw0:
@spdomin Next I'll look into a case that holds the matrix constant and investigate the Muelu preconditioner setup and making sure it doesn't get destroyed. Can you recommend a test to start from?

If you are on the TF sprint, I'll leave this on the fork. Once you're ready to start back in, it might make sense to push a dev branch in NaluCFD.

alanw0:
@spdomin How about the test cvfemHexHC_P3_R0 for looking at heat conduction with the muelu preconditioner? Is there anything else about the input file that might indicate whether it holds the matrix constant or not?

spdomin:
@alanw0, that would be a great test case.

At present, we only have the "recompute_precondiitoner" flag that dictates if we destroy or not. This needs to be done when the connectivity changes. The "reuse_preconditioner" is also an option that keeps the connectivity, however, changes the matrix entries.

In this case, we will want something like:

- matrix__assembly_frequency (or something like this)
    solve_p: 100
    solve_u: 1

The above would say "for the solve_p linear system, update the LHS every 100 steps".

We could also embed this into the solver parameters for each solver block, e.g., solve_p. The issue is that solve_p probably does not know about time stepping. It need to be told what to do.

Let's chat about this through email given our work schedule today.

alanw0:
@spdomin I may need a little help navigating the input file. e.g., in the cvfemHexHC_P3_R0.i file, there is no solve_p and solve_u to associate the matrix_assembly_frequency with. There is a 'temperature: solve_scalar', but really nothing (at least not obvious to me) to indicate that there are two linear-systems being solved at each time step.

alanw0:
@spdomin I take back what I said in the last comment. It looks like there aren't two different linear systems, just two iterations of the non-linear solver each time step. So 'temperature: solver_scalar' is it.

spdomin:
@alan0, I think that the path forward to tie the re-use to the dof is best, as discussed. As such, the above suggestion to use the solver block is obsolete... Let me know via email if you have any questions.

alanw0:
@spdomin ok, I'll let you know. I'm very close to having it working. I got the cvfemHexHC_P3_R0 case working, reusing the lhs and the preconditioner. Now I'm debugging a side-effect issue I seem to have introduced with some of the nightly tests. Should definitely have something for us to discuss next week.

alanw0:
@spdomin I just pushed another commit to this branch, see above. (The message is 'Initial prototype of matrix-lagging working.') It works with the cvfemHexHC_P3_R0 test, and I confirmed with print statements and with the debugger that the preconditioner is only created/computed on the first time-step. This is activated by putting this in the input file in the equation_systems scope:
  matrix_assembly_frequency:
    temperature: 100

Browse through this commit if you like, and/or we can discuss further next week.

alanw0:
Closing this pull request. Will potentially open a new one with cleaned up commit history etc.

PR #235, opened 2017-11-07
Addition of Kokkos View function calls to Quad9 master element class

overfelt:
This continues and hopefully completes retrofitting the master element classes with SharedMemView function signatures.


alanw0:
Howdy James, It looks like this can't merge due to conflicts. I think I have a recipe for you to fix that, and also squash out your intermediate commits such as the one that created a std::vector which later got turned into a scalar.
First update NaluCFD:master into the master branch of your fork, which should go smoothly with no conflicts.
Then you can see the commits you've added in your quad9 branch by going onto your quad9 branch and doing % git log --oneline master..quad9
Then you can do % git rebase -i master
and it will allow you to squash your commits together and produce a single commit message etc. As you finish that it should also give you a chance to resolve any conflicts.
Then you push your quad9 branch to your fork using % git push -f
and it should replace your individual commits with the new one, and then I think your pull request should be able to be rebased onto NaluCFD:master without conflicts.

PR #236, opened 2017-11-07
Higher order implementation for actuator line

gantech:
- Implemented higher order kernels for the RHS term in actuator line. LHS will always be lumped.
- Implemented a unit test for the kernel.
- Changed the spreading of actuator forces for the RHS term to spread directly to the nodes instead of spreading to the element centers and then to the nodes.
- Fixed a bug in the ordering of coordinates passed to the meSCS->IsInElement call. Should've been (x1,x2,x3,...y1,y2,y3,....z1,z2,z3...) instead of (x1,y1,z1,x2,y2,z2,...).
- Reblessed gold norms for the two actuator line tests.

Thanks to @rcknaus for the discussions involved in designing and implementing the higher order kernels and changing the spreading function.


src/MomentumActuatorSrcElemKernel.C
```diff
@@ -0,0 +1,109 @@
+/*------------------------------------------------------------------------*/
```
rcknaus: This looks good.  It would be nice if it was more generic, like taking two vector-type fields instead of specifically the actuator field.

gantech: Cool... so the names of the fields would be inputs to the constructor?

gantech:
All tests pass now.

gantech:
Should I resubmit this pull request with all my changes collapsed into one commit?

sayerhs:
@gantech I don't think that would be necessary, the person doing the merge can choose `squash and merge`... unless you were thinking of rebasing against the current master.

gantech:
So.... I recently merged with master in a10a598. Would a squash and merge affect this?

PR #239, opened 2017-11-08
Start setting up for unit testing nonconformal algorithms

rcknaus:
* Adds a class that constructs two side-by-side element blocks of possibly different topologies with an interface between them


rcknaus:
On second thought think I should hold off on this

spdomin:
Okay, your choice. Code looks good. We can chat tomorrow.

PR #245, opened 2017-11-10
Fix declaration of realmin to remove duplication and use numeric_limits

alanw0:


alanw0:
@sayerhs currently there is no master-element namespace, although it would be easy to create one. Should it just be called sierra::nalu::ME?

sayerhs:
@alanw0 may be `sierra::nalu::MEimpl`. Since this is just an implementation detail. What is the best practice in STK for such things?

alanw0:
@sayerhs yes in stk there is a impl namespace. But stk is a library and the impl namespace encloses classes/functions which aren't really intended for public use. The paradigm in an app like nalu is a little different. We're not really trying to keep things from public use, but it might be nice to segregate subsystems into separate namespaces. If we want to put ME stuff into a separate namespace we should put all the master-elements into the namespace. But that starts to feel like a larger task...
Alternatively, we could make a 'constants' namespace or something similar, to just put realmin into (and possibly other stuff).

alanw0:
@sayerhs I added a commit that does the compromise of moving the realmin constant into a 'MEconstants' namespace. Let me know if you think that's ok. Then at another time we could think about moving all master-elements into a new namespace etc.

sayerhs:
@alanw0 Yeah, that seems like good compromise to me. Thanks for taking the suggestion seriously.

spdomin:
Many thanks... looks great.

PR #252, opened 2017-11-22
Improvement to Hybrid DG/CVFEM scheme with sliding mesh

spdomin:
* remove VRTM in continuity in favor of velocity;
given mesh motion, mesh velocity applies only on current

* remove exposed area vector parallel communication for DG; not required
now that we use the opposing normal calc (side change to clean up
the code base)

* rebless moving mesh DG tests; norms always lower

Ready to push...


spdomin:
@mbarone81 and @overfelt, many thanks for the reviews today:)

PR #253, opened 2017-11-24
Add ke NSO for Z and h

spdomin:
* Added kernel-based NSO with ke fine scale residual

Note:
a) Developed this on my mac. As such, I will modify a regression test ASAP


spdomin:
Yes, I think that the NSO as a turbulence model is covered...

PR #255, opened 2017-11-29
Add a clipped, scaled scalar variance for table lookups

rcknaus:


spdomin:
@rcknaus, go ahead and push the mods for the rest directly...

PR #256, opened 2017-11-30
Allow different tolerance levels for predictor and final corrector iters

sayerhs:
For ABL simulations, we have been using different tolerance settings for the linear solvers, i.e., use a lax tolerance normally except for the final outer iteration when a tighter tolerance level is chosen. This pull request implements the necessary modifications to allow this. 

```
linear_solvers:

  - name: solve_scalar
    type: tpetra
    method: gmres
    preconditioner: sgs
    tolerance: 1e-3
    final_tolerance: 1e-6
    max_iterations: 75
    kspace: 75
    output_level: 0

  - name: solve_cont
    type: tpetra
    method: gmres
    preconditioner: muelu
    tolerance: 1e-3
    final_tolerance: 1e-6
    max_iterations: 75
    kspace: 75
    output_level: 0
    recompute_preconditioner: no
```

Preserves backward compatibility by default to `tolerance` if `final_tolerance` is not provided in the input file.

### DIFF
--- a/include/LinearSolver.h
+++ b/include/LinearSolver.h
@@ -105,7 +105,8 @@ class TpetraLinearSolver : public LinearSolver
     int solve(
       Teuchos::RCP<LinSys::Vector> sln,
       int & iterationCount,
-      double & scaledResidual);
+      double & scaledResidual,
+      bool isFinalOuterIter);
 
     virtual PetraType getType() { return PT_TPETRA; }
     TpetraLinearSolverConfig *getConfig() { return config_; }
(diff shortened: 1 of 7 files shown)


spdomin:
I think that this code is fine. In general, there is not a notion of a predictor and corrector step in Nalu. The scheme is based on a linearized operator split methodology that is repaired by outer iteration. This is in contrast to a classic predictor followed by corrector methodology (for example, Adams Bashforth explicit predictor followed by implicit corrector). Moreover, there is not a PISO-like notion of predict and correct. Therefore, in most cases the code should be run by setting a nonlinear iteration number and a nonlinear residual. 

At present, the code will determine the final step by assuming that the last nonlinear iteration is the step in which the final tolerance is used. However, if the code converges at step 3 with a max iterations of 10, then this parameter will never be used.

I guess my only end point is that there is not a notion of predict/correct. If the team is running simulations with specified nonlinear iterations that are always expected to be run, then this implementation can make sense.

PR #258, opened 2017-11-30
Realm: Add input option to skip parts from Exodus file

sayerhs:
Introduces `skip_parts_from_file` option in Realm section of Nalu input file
that accepts a list of Exodus parts that must be added to Realm's
inactive_selector.

Usage:

```
realms:

  - name: fluids_realm
    mesh: abl_40.g
    use_edges: yes
    automatic_decomposition_type: rcb
    skip_parts_from_file:
      - zplane_0090.0
```

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -556,6 +556,8 @@ class Realm {
     const TurbulenceModelConstant turbModelEnum);
   bool process_adaptivity();
 
+  void process_skipped_parts();
+
   // element promotion options
   bool doPromotion_; // conto
   unsigned promotionOrder_;
@@ -565,6 +567,10 @@ class Realm {
 
   // save off the node
   const YAML::Node & node_;
+
+  //! User-specified inactive parts in the mesh
+  std::vector<std::string> skippedPartNames_;
+  stk::mesh::PartVector skippedParts_;
 
   // tools
   std::unique_ptr<ElementDescription> desc_; // holds topo info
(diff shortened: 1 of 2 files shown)


sayerhs:
@spdomin We could use `stk::mesh::MetaData::get_parts()` to query all parts known to STK and then have populate `Realm::activeParts_` to automate the determination of _inactive parts_. However, we will still need to explicitly add in inactive part for Overset to handle selectors in Algorithms correctly. 

```
  // provide inactive Overset part that excludes background surface
  stk::mesh::Selector inactiveOverSetSelector =
    (hasOverset_) ? oversetManager_->get_inactive_selector()
    : stk::mesh::Selector();

  auto inactivePartsSelector = ( stk::mesh::selectUnion(meta.get_parts() 
                               & !(stk::mesh::selectUnion(activeParts_) );

  return (inactivePartsSelector | inactiveOversetSelector);
```

The annoyance with this is that the selectors do `selectUnion(activeParts)` twice... once with the `partVec_` in algorithms and once in the inactive_selector. I am not sure if that is a big performance implication. @alanw0  comments?

sayerhs:
@spdomin It appears that we already do have `Realm::bcPartVec_`, so all we need is to populate `interiorPartVec_` from `Realm::register_interior_algorithm` and we will have all parts that should be considered _active_.

spdomin:
@shreyas. Sounds great. Personally, I am still trying to sort though the best path forward. If you have ideas, let me know. I am still thinking that we need the overset part to intersect with active. Do you think that we can keep all of the algorithm selectors as they are currently and simply modify what is inactive?

shreyas:
Hey Stefan, I think you have added the wrong shreyas here.

sayerhs:
Closing this request in favor of #263

PR #259, opened 2017-12-01
F/grad bc

mchurchf:
This is a pull request for the feature to specify a boundary normal gradient of temperature as an option with the symmetry boundary conditions.  I have included the code that makes this feature work.  I also updated the ablForcingEdge regression test to use this feature and be our standard neutral ABL example case.


src/NaluParsing.C
```diff
@@ -1144,7 +1144,13 @@ namespace YAML
   bool convert<sierra::nalu::SymmetryUserData>::decode(const Node& node,
     sierra::nalu::SymmetryUserData& symmetryData)
   {
-    // nothing as of yet
+    // This allows the user to set a fixed temperature gradient that is
+    // achieved through application of a compatible heat flux. 
+    if (node["temperature_gradient"])
+    {
+      symmetryData.temperatureGradient_ = node["temperature_gradient"].as<sierra::nalu::NormalHeatFlux>();
```
spdomin: As the intent of the bc is to extract a double, we can simply use the double type here as apposed to a new NormalHeatFlux() type. Either way is fine; this is very clear. However, if we had a special type for every double, the parsing code would substantially grow. Your choice as to whether or not the type safety is required.

src/AssembleScalarNormalGradientBCSolverAlgorithm.C
```diff
@@ -0,0 +1,218 @@
...
+        }
+
+        const int offset = ip*nDim;
+        double areaNorm = 0.0;
+        for (int idir = 0; idir < nDim; ++idir)
+          areaNorm += areaVec[offset+idir]*areaVec[offset+idir];
+        areaNorm = std::sqrt(areaNorm);
+
+        double qn = eviscBip*specHeatBip*gradBip;
+        NaluEnv::self().naluOutputP0() << "Computed q_normal on face node = " << qn << "; specificHeat = " << specHeatBip << "; eviscBip = " << eviscBip << "; gradT = " << gradBip << std::endl;
```
spdomin: This needs to be removed.

src/AssembleScalarNormalGradientBCSolverAlgorithm.C
```diff
@@ -0,0 +1,218 @@
...
+//==========================================================================
+// AssembleScalarNormalGradientBCSolverAlgorithm - 
+//   This boundary conditions applies a scalar normal flux bc, 
+//   Int bcScalarQ*area, such that a specified normal temperature gradient
+//   is achieved.
+//==========================================================================
+//--------------------------------------------------------------------------
+//-------- constructor -----------------------------------------------------
+//--------------------------------------------------------------------------
+AssembleScalarNormalGradientBCSolverAlgorithm::AssembleScalarNormalGradientBCSolverAlgorithm(
```
spdomin: At first, when I thought that the implementation was going to be a flux-based approach, I felt that a general name would suffice. However, after our first meeting, it was very clear that this class can only be used for a specific equation system, e.g., enthalpy. This is due to the specific heat scaling. As such, I would change the name to tie in the EnthalpyEQS somehow. Again, a very minor detail.

reg_tests/test_files/ablForcingEdge/README
```diff
@@ -0,0 +1,68 @@
+Neutral Atmospheric Boundary Layer Case
```
spdomin: README files are a nice addition:) In other projects, we found that sometimes they become out of date and, as such, the input file is the best description. I am not advocating that approach, rather noting that we will not have the responsibility to keep this up-to-date. Again, nothing to change, only something to share with the team.

src/NaluParsing.C
sayerhs: @spdomin Considering that the first thing that is done is extracting the 1 double out of the special struct and passed around as double it doesn't make sense to have a special data type for the user data. Same can be said of roughnessHeight etc.

src/AssembleScalarNormalGradientBCSolverAlgorithm.C
sayerhs: I second this. This algorithm is specific to EnthalpyEQS.

reg_tests/CTestList.cmake
```diff
@@ -66,7 +66,7 @@ endfunction(add_test_r_cat)
 # Regression tests
 #=============================================================================
 
-add_test_r_cat(ablForcingEdge 4 11)
+add_test_r_cat(ablForcingEdge 8 11)
```
sayerhs: @mchurchf I feel that it would be better to rename this test to `ablNeutralEdge` since the scope of this has expanded beyond just an example of the ABL forcing algorithm. It would be more intuitive for new users and a natural point to start from, especially if they are SOWFA users moving over to Nalu codebase.

reg_tests/test_files/ablForcingEdge/ablForcingEdge.i
```diff
@@ -3,58 +3,80 @@ Simulations:
...
     muelu_xml_file_name: ../../xml/milestone.xml
 
+
+# Specify the differnt physics realms.  Here, we just have one for the fluid.
 realms:
 
+  # The fluid realm that uses the 5 km x 5 km x 1 km atmospheric LES mesh.
   - name: fluidRealm
-    mesh: ../../mesh/abl_1km_cube_sample.g
+    mesh: ../../mesh/abl_5km_5km_1km_neutral.g
```
sayerhs: It appears that the submodule hasn't been updated, so merging this will break this test as the mesh will not be found in the NaluMesh submodule commit that is checked out currently. Please add a commit to update the version of submodule

src/NaluParsing.C
spdomin: @sayerhs, I agree; let's have him change this to a double - along with the other new parsing type. This will make things simpler in the end and reduce code.

mchurchf:
I did the following:
1) Removed the commented out print out line. 
2) Renamed from AssembleScalarNormalGradientBCSolverAlgorithm to AssembleTemperatureNormalGradientBCSolverAlgorithm.  
3) Shreyas helped me properly add in the NaluMesh additional mesh.  
I think this is all good to go

PR #262, opened 2017-12-04
Allow different tolerance settings for final/non-final outer iters

sayerhs:
Add an option in `linear_solvers` section to allow user to specify
linear solver convergence tolerance settings for the final and
prior outer iterations.

This commit fixes the error seen in certain regression tests after a
previous attempt at implementing this feature in pull request #256. Please
see the discussion in #256 for @spdomin's admonitions on using this feature,
particularly its interactions with the convergence tolerance setting in EQS
inputs.

### DIFF
--- a/include/LinearSolver.h
+++ b/include/LinearSolver.h
@@ -105,7 +105,8 @@ class TpetraLinearSolver : public LinearSolver
     int solve(
       Teuchos::RCP<LinSys::Vector> sln,
       int & iterationCount,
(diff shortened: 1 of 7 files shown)


sayerhs:
@jrood-nrel I have verified that all tests pass on Peregrine. 

I have also created an issue on Trilinos (trilinos/Trilinos#2040) that discusses the error thrown by certain Belos solvers.

PR #263, opened 2017-12-04
Realm: Automate determination of inactive parts.

sayerhs:
Implement logic to determine inactive parts (i.e., parts in the
Exodus-II database) that aren't explicitly listed in interior or BC.
The nodes from these parts will be ignored during the setup of
TpetraLinearSystem. Realm::get_inactive_selector must still handle
overset separately because certain interior regions can be tagged
inactive by overset simulations.

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -508,6 +508,9 @@ class Realm {
   // beginning wall time
   double wallTimeStart_;
 
+  // mesh parts for all interior domains
+  stk::mesh::PartVector interiorPartVec_;
+
   // mesh parts for all boundary conditions
   stk::mesh::PartVector bcPartVec_;
 
--- a/src/Realm.C
+++ b/src/Realm.C
@@ -2988,6 +2988,9 @@ Realm::register_interior_algorithm(
   else {
     it->second->partVec_.push_back(part);
   }
+
+  // Track parts that are registered to interior algorithms
+  interiorPartVec_.push_back(part);
 }
 
 //--------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


sayerhs:
@spdomin @rcknaus @alanw0 This is an alternative to #258 using interior and BC parts already provided by user to remove additional parts that may be found in the Exodus-II file (from data probes, ABL forcing, etc.) but should not participate in TpetraLinearSystem and or Algorithm nodal/element loops.

rcknaus:
Isn't it possible just to keep the active parts up to date?  As in, if the overset part isn't active, then just make sure it's not in the active parts list.  The inactive parts then is just the complement of the active parts.

spdomin:
I agree with @rcknaus. We discusses this a bit in the TF team room. There may be an easy way to combine the bcPartVec_ and interiorPartVec_ to make the entire code base more clean. We should discuss this over the phone since there may be nuances..

sayerhs:
@rcknaus @spdomin Feel free to open another pull request with your code on how this feature must be implemented. As long as the goal (existence of random parts in an Exodus-II database doesn't break TpetraLinearSolver) is achieved, I am not particular in how that is implemented in the code.

spdomin:
@sayerhs, looks great. Nice that the user need not be aware of the inactive parts.

@michaelasprague, please add a long term story for inactive part design improvement that might look towards some of @rcknaus suggestions above.

michaelasprague:
@spdomin - created story ADSE07-275.

PR #264, opened 2017-12-04
ABL: Add resolved and SFS stress computations for velocity and temperature fields

sayerhs:
This pull request adds ability to compute time-averaged resolved and SFS stress fields for both temperature and velocity fields for use with ABL precursor simulations. This will be used with boundary layer statistics post-processing to extract various quantities of interest during wind energy simulations.

### DIFF
--- a/docs/source/user/nalu_run/nalu_inp.rst
+++ b/docs/source/user/nalu_run/nalu_inp.rst
@@ -944,7 +944,10 @@ Turbulence averaging
    .. code-block:: yaml
 
       turbulence_averaging:
+        forced_reset: no
         time_filter_interval: 100000.0
+
+        averaging_type: nalu_classic/moving_exponential
 
         specifications:
 
@@ -959,6 +962,10 @@ Turbulence averaging
 
             compute_tke: yes
             compute_reynolds_stress: yes
+            compute_resolved_stress: yes
+            compute_temperature_resolved_flux: yes
+            compute_sfs_stress: yes
+            compute_temperature_sfs_flux: yes
             compute_q_criterion: yes
             compute_vorticity: yes
             compute_lambda_ci: yes
@@ -969,13 +976,32 @@ Turbulence averaging
    prefixed with ``turbulence_averaging.name`` but only the variable
    name after the period should appear in the input file.
 
+.. inpfile:: turbulence_averaging.forced_reset
+
+   A boolean flag indicating whether the averaging of all quantities in the
+   turbulence averaging section is reset. If this flag is true, the
+   running average is set to zero.
+
+.. inpfile:: turbulence_averaging.averaging_type
+
+   This parameter sets the choice of the running average type. Possible
+   values are:
+
+   ``nalu_classic``
+     "Sawtooth" average. The running average is set to zero each time the time
+     filter width is reached and a new average is calculated for the next time
+     interval.
+
+   ``moving_exponential``
+     "Moving window" average where the window size is set to to the time
+     filter width. The contribution of any quantity before the moving window
+     towards the average value reduces exponentially with every time step.
+   
 .. inpfile:: turbulence_averaging.time_filter_interval
 
-   Number indicating the time filter size over which calculate the
-   running average. The current implementation of the running average
-   in Nalu uses a "sawtooth" average. The running average is set to
-   zero each time the time filter width is reached and a new average
-   is calculated for the next time interval.
+   Number indicating the time filter size over which to calculate the
+   running average. This quantity is used in different ways for each filter
+   discussed above.
 
 .. inpfile:: turbulence_averaging.specifications
 
@@ -1007,6 +1033,36 @@ Turbulence averaging
    A boolean flag indicating whether the reynolds stress is
    computed. The default value is ``no``.
 
+.. inpfile:: turbulence_averaging.specifications.compute_resolved_stress
+
+   A boolean flag indicating whether the resolved stress is
+   computed. The default value is ``no``. When this option is turned on, the
+   Favre averages of the resolved quantities are computed as well.
+   
+.. inpfile:: turbulence_averaging.specifications.compute_temperature_resolved_flux
+
+   A boolean flag indicating whether the resolved temperature flux is
+   computed. The default value is ``no``. When this option is turned on, the
+   Favre averages of the resolved temperature is computed as well.
+
+.. inpfile:: turbulence_averaging.specifications.compute_sfs_stress
+
+   A boolean flag indicating whether the sub-filter scale stress is
+   computed. The default value is ``no``. The sub-filter scale stress model is
+   assumed to be of an eddy viscosity type and the turbulent viscosity computed
+   by the turbulence model is used. The sub-filter scale kinetic energy is used
+   to determine the isotropic component of the sub-filter stress. As described
+   in the section :ref:`supp_eqn_set_mom_cons`, the Yoshisawa model is used to
+   compute the sub-filter kinetic energy when it is not transported. 
+   
+.. inpfile:: turbulence_averaging.specifications.compute_temperature_sfs_flux
+
+   A boolean flag indicating whether the sub-filter scale flux of temperature is
+   computed. The default value is ``no``. The sub-filter scale stress model is
+   assumed to be of an eddy diffusivity type and the turbulent diffusivity computed
+   by the turbulence model is used along with a constant turbulent Prandtl number
+   obtained from the Realm.
+   
 .. inpfile:: turbulence_averaging.specifications.compute_favre_stress
 
    A boolean flag indicating whether the Favre stress is computed. The
--- a/docs/source/user/nalu_run/nalu_inp.rst
+++ b/docs/source/user/nalu_run/nalu_inp.rst
@@ -1035,33 +1035,37 @@ Turbulence averaging
 
 .. inpfile:: turbulence_averaging.specifications.compute_resolved_stress
 
-   A boolean flag indicating whether the resolved stress is
-   computed. The default value is ``no``. When this option is turned on, the
-   Favre averages of the resolved quantities are computed as well.
+   A boolean flag indicating whether the average resolved stress is 
+   computed as :math:`< \bar\rho \widetilde{u_i} \widetilde{u_j} >`.
+   The default value is ``no``. When this option is turned on, the Favre
+   average of the resolved velocity, :math:`< \bar{\rho} \widetilde{u_j} >`, is
+   computed as well.
    
 .. inpfile:: turbulence_averaging.specifications.compute_temperature_resolved_flux
 
-   A boolean flag indicating whether the resolved temperature flux is
-   computed. The default value is ``no``. When this option is turned on, the
-   Favre averages of the resolved temperature is computed as well.
+   A boolean flag indicating whether the average resolved temperature flux is
+   computed as :math:`< \bar\rho \widetilde{u_i} \widetilde{\theta} >`. The
+   default value is ``no``. When this option is turned on, the Favre average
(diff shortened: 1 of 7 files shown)


src/TurbulenceAveragingPostProcessing.C
```diff
@@ -735,6 +869,233 @@ TurbulenceAveragingPostProcessing::compute_favre_stress(
...
+
+    // Store the xx, xy, xz, yy, yz, zz components of sfs_stress
+    for ( stk::mesh::Bucket::size_type k = 0 ; k < length ; ++k ) {
+      double divU = 0.0;
+      for ( int j = 0; j < nDim; ++j)
+          divU += dudx_[k*offSet+(nDim*j+j)] ;
+      double sfstke = 0.0;
+      if(computeSFSTKE) {
+          // Use method of Yoshisawa (1986) - Statistical theory for compressible turbulent shear flows, with the application to subgrid modeling, 29, 2152.
+          double Ci = 0.9;
```
rcknaus: Probably shouldn't just be a hard coded value.  One option would be to just put Ci into the turbulent constants map.  Another would be to compute an estimate of Ci from the values of cmuCs and cmuEps already in the turbulent constants map ( ci = cs^4/cnu^2 ).  I'd prefer just putting Ci into the constants map.

src/TurbulenceAveragingPostProcessing.C
```diff
@@ -735,6 +869,233 @@ TurbulenceAveragingPostProcessing::compute_favre_stress(
...
+          double sijMag = 0.0;
+          for ( int i = 0; i < nDim; ++i ) {
+              for ( int j = 0; j < nDim; ++j ) {
+                  const double rateOfStrain = 0.5*(dudx_[k*offSet+nDim*i+j] + dudx_[k*offSet+nDim*j+i]);
+                  sijMag += rateOfStrain*rateOfStrain;
+              }
+          }
+          sijMag = std::sqrt(2.0*sijMag);
+          sfstke = 2.0 * Ci * std::pow(dualNodalVolume_[k], 2.0*invNdim) * sijMag;
```
rcknaus: Ci * std::pow(dualNodalVolume_[k], 2.0*invNdim) * sijMagSq . 

Alternatively, you can model the trace of the subgrid tensor, tau_kk = 2 rho ksgs = 2 Ci delta^2 rho |S|^2

src/TurbulenceAveragingPostProcessing.C
```diff
@@ -735,6 +869,233 @@ TurbulenceAveragingPostProcessing::compute_favre_stress(
...
+      } else {
+          sfstke = turbKe_[k];
+      }
+      size_t componentCount = 0;
+      for ( int i = 0; i < nDim; ++i ) {
+          for ( int j = i; j < nDim; ++j ) {
+              const double divUTerm = ( i == j ) ? 2.0/3.0*divU : 0.0;
+              const double sfsTKEterm = ( i == j ) ? 2.0/3.0*sfstke : 0.0;
```
rcknaus: 2.0 / 3.0 * rho * ksgs

src/TurbulenceAveragingPostProcessing.C
```diff
@@ -735,6 +869,233 @@ TurbulenceAveragingPostProcessing::compute_favre_stress(
...
+              componentCount++;
+          }
+      }
+    }
+  }
+}
+
+
+void
+TurbulenceAveragingPostProcessing::compute_temperature_sfs_stress(
```
rcknaus: Maybe call this the``sfs temperature flux"?  It's not a stress, really.  Same with the resolved portion.

sayerhs:
@gantech can you address @rcknaus comments?

gantech:
Yup..I'll work on those. Thanks @rcknaus

src/TurbulenceAveragingPostProcessing.C
sayerhs: @gantech If this is being renamed, also rename the input file options.

src/TurbulenceAveragingPostProcessing.C
spdomin: Many options here... msgs is best, however, may not be transported.

src/TurbulenceAveragingPostProcessing.C
spdomin: Is SFS a term used in atmospheric sciences? In general, I would like to see a generalized approach so that we could compute, for example, any \bar{rho*u'\phi'}. Would that be too hard? Otherwise, we will need to generalize a bit later. I could use this capability in a variety of projects...

sayerhs:
Closing pending new pull request rules.

michaelasprague:
This is an important feature for precursor simulations; let's continue the review.

spdomin:
If the correctness of the implementation is addressed, I am fine with this incremental push with the defined process being to add a generalization task to the backlog. I will let @rcknaus drive the correctness questions - which may have already been addressed.

sayerhs:
@spdomin - @rcknaus's comments are being addressed by @gantech. This will be merged only after Ganesh's commits are in.

spdomin:
@sayerhs, does that mean someone is pushing code that we know is wrong, or is there healthy debate as to the model form?

gantech:
We had a preliminary discussion of the relevant statistics that are necessary to compare ABL simulations for the Q4 milestone on HivLo. @rcknaus and I had written up a document regarding the same and had uploaded it to Jira. After that, I had implemented the computation of sfs stresses in consultation with @rcknaus. I had some mistakes in the computation of the sfs tke when a transport equation for the same wasn't being used. I have addressed those comments yesterday and was making sure that all regression tests pass.

src/TurbulenceAveragingPostProcessing.C
gantech: Thanks @spdomin. The code does check if the variable "turbulent_ke" exists and uses that if available. If "turbulent_ke" is not available, then the sfs tke is computed using the Yoshisawa model.

src/TurbulenceAveragingPostProcessing.C
gantech: Thanks @rcknaus. I moved this to the turbulent constants map.

src/TurbulenceAveragingPostProcessing.C
gantech: Thanks for catching this error @rcknaus. I have addressed these issues. We're cross checking before pushing it here.

sayerhs:
Commit a30bfe4673f addresses @rcknaus' request for changes.

michaelasprague:
Per standup discussion, please provide appropriate set of documentation.

mbarone81:
@gantech @rcknaus I am trying to locate the writeup posted to Jira that Ganesh references in this comment stream but cannot seem to find it.  Can someone please point me to it?

sayerhs:
@mbarone81 The milestone related documents are here https://hfm-exawind.atlassian.net/browse/A2EHFM-5

gantech:
Thanks @sayerhs for pointing this out.

mbarone81:
Hi @gantech Which document contains the definition of the statistical quantities addressed in this pull request?  Are the formulae for these written down?  I can't find them in the quarterly milestone docs.  I ask because there can be some ambiguity when defining some of these quantities.

gantech:
@mbarone81 

The SFS stresses are as defined in http://nalu.readthedocs.io/en/latest/source/theory/supportedEquationSet.html#equation-deviatoric-stress-les. I point to this section in the documentation through this commit https://github.com/NaluCFD/Nalu/pull/264/commits/eb0c2d884d39de857a1cad6e62e97e7d3ec601ff. 

In addition the resolved stress and resolved temperature flux are computed as follows:

Resolved stress: rho * u_i * u_j, 
Resolved temperature flux: rho * u_i * theta.

mbarone81:
OK, thanks, I will work off these theory doc definitions...

docs/source/user/nalu_run/nalu_inp.rst
```diff
@@ -1007,6 +1033,36 @@ Turbulence averaging
...
+   Favre averages of the resolved temperature is computed as well.
+
+.. inpfile:: turbulence_averaging.specifications.compute_sfs_stress
+
+   A boolean flag indicating whether the sub-filter scale stress is
+   computed. The default value is ``no``. The sub-filter scale stress model is
+   assumed to be of an eddy viscosity type and the turbulent viscosity computed
+   by the turbulence model is used. The sub-filter scale kinetic energy is used
+   to determine the isotropic component of the sub-filter stress. As described
+   in the section :ref:`supp_eqn_set_mom_cons`, the Yoshisawa model is used to
```
mbarone81: Is this documentation included in the pull request?

gantech: Yes. This was in the last commit https://github.com/NaluCFD/Nalu/commit/eb0c2d884d39de857a1cad6e62e97e7d3ec601ff after the discussion during the standup last Thursday.

mbarone81:
The SFS quantities look OK.  Two potential issues with the resolved stress.  1) The term 'resolved stress' is often formally defined using a double-filtering operation (see, e.g., Pope's book).  This notion of the resolved stress is used in dynamic sgs models.  If we keep this terminology, we should have an equation in the user manual that explicitly defines what is meant by 'resolved stress.' 2) The implementation is computing <u_i u_j>.  Do we anticipate having a use for this particular quantity? Usually we are using <u'_i u'_j>, which is currently covered in the code by the Reynolds averaging operator.   Finally, I see  Yoshizawa's formula oin the theory manual...just need to correct spelling of his name in both places :-)

mbarone81:
I see that the resolved_stress is actually being computed by <rho u_i u_j> so it is the current Favre stress operator in the code that is more relevant, and should give <rho u'_i u'_j>/<rho>.

gantech:
Thanks Matt. I do add add the density through the commit https://github.com/NaluCFD/Nalu/pull/264/commits/cb502c9ad43e4bbfc419fadc29ab14bf5b1d2066 as well. The purpose of computing the resolved stress is to later compute <u_i' u_j'> through manipulation of Favre averaged <u_i u_j> and <u_i>. This was to be covered through pull request https://github.com/NaluCFD/Nalu/pull/265. 

If the computation of the resolved stress seems ok, then I can add a few more comments in the documentation for the resolved stress and correct the spelling of Yoshikawa through another commit.

mbarone81:
OK, so long as it is made clear in the docs what is being output then that is fine by me.

gantech:
Ok thanks.... I'll add another commit to clarify the exact resolved stress computed through this option in `turbulence_averaging`. Also will fix the Yoshizawa spelling.

spdomin:
In my eigenvalue decomposition methods, I need all of these quantities as I determine eigenvalues for the resolved and modeled terms. 

In general, I have moved away from nodal-based approaches for the modeled terms as this drives the need for a design-order projected nodal gradient for the higher-order schemes. As such, I have moved to use elemental values with quadrature over the element. This provides and elemental field that is design-order. However, when this is the case, the line-of-site becomes problematic since it only works for nodes.

This is just a detail/caveat should someone start looking at these modeled terms at the nodes when P>1 is in use. As noted, activation of a PNG fixes this at the expense of solving a new system.

mchurchf:
I think you should merge this.  I think it is a solid incremental pull request that we can start using immediately for ABL testing.

gantech:
Thanks @mchurchf. I'm running the tests right now on the rebased version that we will directly push to master citing this pull request. Should be ready to go in about 1 hour.

sayerhs:
Based on the discussion during standups the code discussed and approved in this pull request was rebased, squashed and committed in e0fd57241.

spdomin:
If someone could add the comments above to the follow-on clean-up story, that would be great. Specifically, the higher-order interface with nodal post-processing.

PR #266, opened 2017-12-04
Deprecate non-consolidated element-based PMR in favor of SIMD/Kokkos

spdomin:
* addded LHS = ((abs+scattering)*I - radSrc)*scv kernel; LMM hooked in
	name: absorption_black_body

* added LHS = (-G*scattering/4/pi)*scv kernel; LMM || LMM hooked in
	name: isotropic_scattering

* add LHS = sj*dI/dxj*njdS + SUCV kernel
	name: advection_sucv

* removed AssembleElemSolverAlgorithmDep as no one uses this now - hurray!

* removed FEM SuppAlgs as this is not a really useful production capability
as of yet. The non-SUCV CVFEM and SUPG was odd- kind of reseach path. In the end,
SUPG and SUCV performed the same. Moreover, the pure FEM-based kernel never had
the proper BC activated so that a pure FEM-based solve could be tested. We can
manage that later should PMR project paths push us down there.

* add "intensity" to eqn name so that kernels can be found and built.

* modified concentricRad and inputFireElem to use new consolidated. concentricRad
is within 1e-14 while inputFireElem changed from SUP to SUCV. However, SUCV
showed a lower norm and a bit more monotonicity.

Notes:

a) consolidated is fully active for PMR. Ready for S-sync.
b) ran a number of flows, e.g., fluidsCHTPMR with comparison to edge- and elem-cons.
All looks great.


spdomin:
Messed up the copyright...

PR #271, opened 2017-12-07
ABL Forcing

tonyinme:
This is a modification to the ABL forcing capabilities in Nalu.

1. The ABL Forcing momentum source term has been modified. It is now computed by taking the planar average of density and multiplying it by the planar averaged velocity. The temperature source term in the enthalpy equation remains the same.

2. The theory manual has been updated to explain how the ABL Forcing source term works. Some minor modifications to the theory manual have been implemented.

3. The user manual has been updated by adding the ABL Source term capabilities.

4. The norm gold standard file for the ablForcingEdge regression test has been updated.


docs/source/user/nalu_run/abl_forcing.rst
```diff
@@ -0,0 +1,143 @@
...
+  :math:`y` direction.
+
+
+.. inpfile:: abl_forcing.momentum.velocity_z
+
+  A set of lists containing the time in the first element,
+  followed by the desired velocity at each plane in the
+  :math:`z` direction.
+
+.. inpfile:: abl_forcing.temperature.temperature
```
sayerhs: Clarify that temperature section can have its own heights and target parts?

include/MomentumABLForceSrcNodeSuppAlg.h
```diff
@@ -41,8 +41,8 @@ class MomentumABLForceSrcNodeSuppAlg : public SupplementalAlgorithm
   //! Pointer to the dual volume of the mesh
   ScalarFieldType* dualNodalVolume_;
 
-  //! Pointer to the density (state NP1) of the mesh
-  ScalarFieldType* density_;
+  //~ //! Pointer to the density (state NP1) of the mesh
```
sayerhs: Delete unused code

include/ABLForcingAlgorithm.h
```diff
@@ -205,6 +205,9 @@ class ABLForcingAlgorithm
   //! Planar average velocity calculated on the surface [num_UHeights, 3]
   Array2D<double> UmeanCalc_;
 
+  //! Planar average density calculated on the surface [num_THeights]
```
sayerhs: This should be `[num_UHeights]`

tonyinme:
@sayerhs thanks for your comments. I have added these corrections.

src/ABLForcingAlgorithm.C
```diff
@@ -153,6 +154,7 @@ ABLForcingAlgorithm::load_momentum_info(const YAML::Node& node)
   const int ndim = realm_.spatialDimension_;
   if (momSrcType_ == COMPUTED)
     UmeanCalc_.resize(nHeights);
+    rhoMeanCalc_.resize(nHeights);
```
rcknaus: rhoMeanCalc_ isn't in the scope of the if statement here.  Add some braces.

src/ABLForcingAlgorithm.C
```diff
@@ -380,6 +391,7 @@ ABLForcingAlgorithm::create_transfers()
 
   if (momentumForcingOn())
     populate_transfer_data("velocity", velPartNames_);
+    populate_transfer_data("density", velPartNames_);
```
rcknaus: Braces.

tonyinme:
Hey @rcknaus, good catch! I have fixed the if-statements.

docs/source/theory/supportedEquationSet.rst
```diff
@@ -422,6 +402,104 @@ following temperature form:
...
+
+where :math:`\left< \right>` denotes volume averaging at a
+certain time :math:`t`,
+:math:`\mathrm{U}_i` is the desired spatial averaged
+velocity,
+and :math:`\Delta t` is the time-scale between when the source term is computed
+(time :math:`t`) and when it is applied (time :math:`t + \Delta t`).
+This is typically chosen to be the simulation time-step.
+In the case of an ABL simulation with flat terrain, the voulme averaging is done
+over an infinitesimal small plane over the :math:`x` and :math:`y` directions,
```
mbarone81: "over an infinitesimally thin slice in the x and y directions..."

docs/source/theory/supportedEquationSet.rst
```diff
@@ -422,6 +402,104 @@ following temperature form:
...
+where :math:`\theta_{\rm ref}` is the desired spatial averaged temperature,
+:math:`\left< \theta \right>` is the spatial averaged temperature,
+:math:`C_p` is the heat capcity,
+:math:`\alpha_\theta` is the relaxation factor,
+and
+:math:`\Delta t` is the time-scale.
+
+The present implementation can vary the
+source terms as a function of time and space using either a user-defined table
+of previously computed source terms (e.g., from a *precursor* simulation or
```
mbarone81: Is the user_defined forcing implemented?

sayerhs: @mbarone81 Yes, it is implemented, but hasn't seen a lot of testing though. We will have to test it when we get to the inflow/outflow simulations.

mbarone81: Can you point me to it for my own sake?  I missed it....

sayerhs: @mbarone81 take a look at the method `compute_momentum_sources` https://github.com/tonyinme/Nalu/blob/1d97ae17a093dc53aff0cf3616992f9954ac6dc1/src/ABLForcingAlgorithm.C#L451. The `else` block implements `user_defined` in this case the `velocity_x`, `velocity_y` entries in input file are interpreted as source terms instead of a target velocity.

sayerhs: I guess some optimizations could be implemented to skip transfers when the forcing is `user_defined` currently it will do the search without consideration of that flag.

mbarone81: Good point.  It would also be good to alleviate the requirement to have a from_target_part specified when the forcing is user_defined.

sayerhs: Agreed, in general we could minimize the places where we need target parts... by default we should be able to pull interior/material_property parts and use those.

docs/source/theory/supportedEquationSet.rst
tonyinme: Hey @mbarone81, Thanks, I have corrected this.

docs/source/theory/supportedEquationSet.rst
mbarone81: @tonyinme Do you think that modification to skip transfers when the forcing is user_defined is a straightforward task that fits with the resources of the related story?  If so, I recommend doing it for this pull request.  If it is a bigger effort then I recommend that it be captured in a subsequent story.

tonyinme: Hey @mbarone81, I am not sure, it does not look straight forward to me. @sayerhs, would this be easy to do?

sayerhs: @tonyinme My first guess is it should be straightforward with a few `if` conditions, can we sit together tomorrow and look at the code and decide? We can report back at tomorrow's standup. @mbarone81 will that be ok?

mbarone81: Yes, sure, that would be fine.

tonyinme: sounds good to me!

PR #272, opened 2017-12-08
HYPRE Solver integration

sayerhs:
Add HYPRE as an alternative Pressure Poisson solver within Nalu.

    - Introduce two additional method for linear_solver entries

      - tpetra_hypre: Use TpetraLinearSystem for assembly but use Hypre
        solver/preconditioners to solve the resulting linear system.

      - hypre: New HypreLinearSystem for assembly and sumInto directly into
        HYPRE data structures.

    - Currently supported/tested solvers: BoomerAMG, GMRES

    - Currently supported/tested preconditioners: BoomerAMG, Euclid

    - Introduce ENABLE_HYPRE option in CMake to make HYPRE TPL optional; use
      HYPRE_DIR to specify install location during CMake configure phase

    - Add regression test with Hypre as Poisson solver

    - Update documentation with Hypre related input file options

Limitations:

    - The `tpetra_hypre` option requires `activate_aura: on` and is not
      generally as robust as the `hypre` option. However, it is provided in case
      we want to explore mixing Hypre preconditioners with Tpetra GMRES solvers
      in future.


sayerhs:
This pull request adds the Hypre interface that was used by @sthomas61 for the Hypre-MueLU comparison studies as well as the ExaWind FY18-Q1 milestone V27 benchmarking studies. The recommended reviewers are based on suggestions from Stefan, Srini, and Sprague. 

- Stefan/Alan for overall code design and review
- Andrey/Jonathan for feedback regarding whether combining Hypre preconditioners with Belos GMRES solvers is a good idea for future. This was an example in the xSDK Trilinos package
- Srini for clarification around the copyright questions for `XSDKHypreInterface.h` header file
- Jon Rood for `ENABLE_HYPRE` CMake build configuration, documentation updates, introduction of new TPL dependency.

spdomin:
Sorry, I need more time on this:) I feel slow with everyone else's fast okay. As such, I beg patience. 

As a high-level first question, I am confused by: include/XSDKHypreInterface.h. Based on what I have been hearing from @sayerhs, there was a first effort (as a proof-of-concept) to bootstrap Hypre assembly/solve through the xSDK interface. This allowed the team to test quickly to understand the viability of the Hypre approach.

Once this interface proved to be interesting to pursue, I have several correspondences that say that a direct interface to Hypre was desired. This fixed the overhead of copies and also some other nuances with overset/DG schemes in which ghosted entries were problematic.

With all that being said, I do not understand why anything xSDK is part of this push? As such, I need to understand some of the details associated with this push. I can take this offline with SA.

sayerhs:
@spdomin I am using the implementation of `FunctionParameter` from the XSDK header file rather than reinventing the wheel. This interface allows us to take input options in Nalu input file and pass them to Hypre at the appropriate time. I am using this `FunctionParameter` interface (see HypreLinearSolverConfig.C`) even when we are directly populating the Hypre data structures in `HypreLinearSystem.C`. In order to avoid incurring a dependency on xSDKTrilinos, I have copied this header from xSDKTrilinos codebase (originally named `Ifpack2_Hypre.hpp`) and renamed it `XSDKHypreInterface.h` in the Nalu codebase. Srini is discussing with Heroux and xSDKTrilinos team whether this is ok or not.

jrood-nrel:
Would moving to xSDK Trilinos be something we should think about?

sayerhs:
@jrood-nrel I don't see any active development of xSDKTrilinos (https://github.com/trilinos/xSDKTrilinos), so I am not sure of the benefit of adding this on rather than just using one of their header files (which is standalone) directly from Nalu codebase.

sayerhs:
@michaelasprague @spdomin @alanw0 After looking at the code over the weekend, I have decided to remove the code related to the Tpetra->Hypre matrix population from this pull request. Initially, I was hoping to use it to explore the benefits of mixing and matching solvers/preconditioners (i.e., Belos GMRES solver with Hypre BoomerAMG preconditioner). However, the benefits of that are currently unclear and, if we decide to go down that pathway, I would like to design that differently than what I had working via xSDK. 

This change makes the pull request smaller, and addresses Stefan's initial concerns/questions regarding `XSDKHypreInterface.h`. That file is now around 250 lines of code compared to the previous 1300+ lines of code. The code retained deals with providing a high-level interface to pass configuration parameters from input file via HYPRE API calls and initializing different solvers available within Hypre -- see usage in `HypreLinearSolverConfig.C` and `HypreDirectSolver.C`.

sayerhs:
Update: Mike Heroux has indicated that he is fine with the copy of `XSDKHypreInterface.h` file from xSDKTrilinos package into Nalu.

spdomin:
Sorry, I am still working through this off and on:)

Does this implementation work for non-continuity solves? Also, a combined use-case of using Hypre B-AMG and Trilinos Balos/GMRES is not of interest nor or code capability? I think that the answer to these are both "yes", however, wanted to make sure.

sayerhs:
@spdomin 

- The code in the pull-request can be used for non-continuity solves. We have tested it on simple ABL problem. However, we will still need to do some work on getting the proper settings for non-poisson solves. 

- I have removed the extra code that could have enabled mixed solver/preconditioner (boomerAMG + Belos GMRES) because it required maintaining three different maps and added a lot of code complexity.

sayerhs:
@spdomin Added manual HYPRE build instructions as well as instructions to link HYPRE to Nalu for users not using Spack to manage builds.

PR #273, opened 2017-12-08
Fix copyright notices on Kernel files

sayerhs:
Based on discussions with SNL legal, reverting copyright notices on
Kernel files which were originally implemented as SupplementalAlgorithms
with the only changes introduced being conversion to Kokkos based data
structures.


sayerhs:
@spdomin I left `MomentumActuatorSrcElemKernel.{C,h}` files alone because they don't have an equivalent SupplementalAlgorithm version and were created by @gantech.

spdomin:
@sayerhs, looks good. I think that you used a good rule for deciding what was changed and what was not.

PR #276, opened 2017-12-13
Fixed bug which created a compilation warning.

tonyinme:
The order of two variables initialized in the ABL forcing subroutine has been changed in order to avoid a compilation warning.

This is a minor bug fix.


sayerhs:
@tonyinme Can you please create this fix against the latest `master` branch of Nalu and resubmit the pull request, so that there is only one incremental commit and not a copy of all the commits that were squashed before the merge of your other pull request?

PR #284, opened 2017-12-20
Add a Boussinesq buoyancy source term based on a time-averaged ref. temperature

rcknaus:
* Adds (T-Tbar) Boussinesq buoyancy model, where Tbar is computed through a moving average with some specified timescale.  

* Adds a few unit tests for the moving average itself and the boussinesq source term. 

* See https://github.com/NaluCFD/Nalu/pull/175 for more discussion/info.


spdomin:
@rcknaus, thank you for generalized the moving average interface.

PR #285, opened 2017-12-21
Minor typo fix: Grab SDR at state NP1 not N.

marchdf:
Unless there is a good reason for the way it was, I think this is a typo. This minor typo does not affect the unit tests or regressions.


spdomin:
If you can please try and evaluate the SST treat suite to see why the above change did not cause a diff.

marchdf:
There is an SST_DES test in the regression test suite: elemBackStepLRSST. There is no plain SST regression test. In elemBackStepLRSST, only one non-linear iteration is taken. This is why I see no diffs on peregrine when I run the tests.

PR #290, opened 2017-12-29
D/grad bc

mchurchf:
Added documentation to the theory manual and the verification manual about the symmetry BC option to specify the boundary-normal temperature gradient.


sayerhs:
@mchurchf Can you please rename `ablForcingEdge` to `ablNeutralEdge` as it is the most representative example of how we will run ABL and has grown to cover more than just the forcing term, including this temperature gradient BC.

mchurchf:
Addressing Stefan's comments:
- I think I fixed the merge conflict.
- In the previous closed pull request where I added the actual grad T BC feature, I added rtest coverage to the case which was called ableForcingEdge (but this has been renamed to ablNeutralEdge in this pull request)


Addressing Shreyas's comment:
- I renamed the rtest case as you suggest.

PR #292, opened 2018-01-03
Optimization in sumInto, saves ~10% of assembly time

alanw0:
Store a reference to the local-matrix and local-rhs objects
instead of getting them during every call to sumInto. The
reference is refreshed during each call to finalizeLinearSystem
so that it isn't invalidated by mesh motion etc.


spdomin:
@NaluCFD/core - let's hold back on pushes until the nightly tests are clean again. Is that okay? Its always more difficult to separate out new diffs introduced when the repos is diffing.

PR #293, opened 2018-01-03
Added Momentum kernel using a tensor eddy viscosity.

marchdf:
This implements SFS source term in Eq. 7 of the hybrid write-up (where
the SFS stress is defined in Eq. 5). The associated unit test
reproduces the results of the advection-diffusion momentum
kernel (without the advection term) by using a tensor eddy viscosity
which is equivalent to a scalar viscosity (mu on the diagonal and zero
on the off-diagonal).

Documentation of the new source term is provided.


src/MomentumHybridTurbElemKernel.C
```diff
@@ -0,0 +1,209 @@
...
+
+    // save off some offsets
+    const int ilNdim = il * AlgTraits::nDim_;
+    const int irNdim = ir * AlgTraits::nDim_;
+
+    // zero out vector that prevail over all components
+    for (int i = 0; i < AlgTraits::nDim_; ++i) {
+      const int offset = i * AlgTraits::nDim_;
+      for (int j = 0; j < AlgTraits::nDim_; ++j) {
+        w_mutijScs[offset + j] = 0.0;
```
spdomin: You can probably choose to define this as (ndim,nDim) to avoid offsets - either way is fine.

marchdf:
To respond to the high-level questions:
1. Yes, that is the intention. This was modeled on how the NSO is done.
2. Similarly to 1., Is there an open boundary condition implemented for the NSO that I could use as a template or is this a separate functionality needed for the tensor eddy viscosity?
3. The other quantities will be populated with other kernels/algorithms that we are working on at the same time (see the Jira stories).

spdomin:
The NSO assumes zero normal gradient at all bcs. As such, there is no contribution at an open bc.

At present, you would need to mix the new-style interior kernel with an "old-style" AssembleMomElemOpenTVSolverAlg. I suppose I should first look at the model form for this term to see if you zero out the normal gradient, what remains? Do you know? For the standard laminar isotropic viscous stress, a tangential component remains and must be applied. We can talk about this after the next stand-up if you like.

michaelasprague:
@sayerhs @spdomin @marchdf @jamelvin @jrood-nrel  -- looking at the last comment by @spdomin , was this merged prematurely?  If so, what is the best way to remedy? Should we open a relevant issue, or should we revert the pull request?

spdomin:
I suppose that this all about philosophy of code development and the value added to peer reviews. In this case, a unit test was added. As such, this is a high quality path forward for any push. In the main code base, the hook for activating this scheme was provided in MomEQS. However, fields required in the kernel are not yet available. 

The push could have either removed the interface to the kernel in the code or pushed to a development branch. @sayerhs has been advocating development branches for long WIP efforts that have multiple developers contributing. Robert and I used this for higher-order and we also created a development branch for NGP.

We talked about a process for peer review. I think that a +2 or +1 are, respectively, "looks great" and "looks okay" - for commit while 0, -1 and -2 are, respectively, "meh", "hmm, possibly not", and "no way, please". if we follow this with respect for the team's decision, we should be fine.

marchdf:
I can remove the interface to the kernel in the MomEqs so that the user can not access it (I will submit a pull request shortly). @jamelvin and I are actually working on a separate branch for turbulence model development.

Issue #296, opened 2018-01-05
OpenFAST regression test portability


jrood-nrel:
Currently the OpenFAST regression test is not portable to OSX as it expects a `.so` to be loaded:

```
Running ElastoDyn.
 Running AeroDyn.
 Running AirfoilInfo.
 Running BEM.
 Running UnsteadyAero.
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 1, Blade = 1
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 1, Blade = 2
 Warning: Turning off Unsteady Aerodynamics because C_nalpha is 0.  BladeNode = 1, Blade = 3
 Running ServoDyn.
 Running ServoDyn Interface for Bladed Controllers.
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: FAST_InitializeAll:SrvD_Init:BladedInterface_Init:The dynamic library ../../mesh/5MW_Baseline/ServoData/libDISCON_glin64.so could not be loaded. Check that the file exists in the specified location and that it is compiled for 64-bit applications.
```

jrood-nrel:
This is resolved by #299 .

Issue #274, opened 2017-12-12
Allow CMake to search shared libraries with `find_package`


sayerhs:
Currently, Nalu's top-level CMakeLists.txt sets 

```
SET(CMAKE_FIND_LIBRARY_PREFIXES lib)
SET(CMAKE_FIND_LIBRARY_SUFFIXES a)
```

which forces CMake to only look for static libraries with `find_package` directives. Is there a use case where this is explicitly needed and why CMake's platform specific defaults are not appropriate? If this is explicitly needed can we change the second line to 

```
SET(CMAKE_FIND_LIBRARY_SUFFIXES .a .so .dylib)
```

so that it can accept shared libraries if static libraries are unavailable on the user's system? 

Cc: @spdomin @jrood-nrel

jrood-nrel:
I would prefer to go further and remove those two lines and let CMake find any type. Then I would change:

```
find_library(YAML_LIBRARY NAMES libyaml-cpp.a PATHS ${YAML_DIR}/lib )
```
to
```
find_library(YAML_LIBRARY NAMES yaml-cpp PATHS ${YAML_DIR}/lib)
```

jrood-nrel:
I'm happy to address this stuff since this is one of the things I addressed when I was working to reorganize CMakelists a while ago but had to revert because I messed up the location of `PROJECT(Nalu)` which caused problems.

sayerhs:
@jrood-nrel Yes, the preference would be to not explicitly list suffixes in the Nalu CMakeLists.txt file. 

Would also be good to move YAML-cpp stuff into a `cmake/FindYAML.cmake` along with package helpers.

spdomin:
Dealing with finding yaml would be a good change:)

jrood-nrel:
This is resolved by #299 .

PR #299, opened 2018-01-10
Making OpenFAST test more portable and allowing for linking against shared libraries

jrood-nrel:
This pull request addresses #296 and #274 .

For #296 this moves the servo library and other necessary files for the OpenFAST test from the `mesh` directory into the test directory itself, and more importantly the `CMAKE_BINARY_DIR`, so that the file output generated from OpenFAST itself does not dirty the `mesh` directory or the Nalu source directory. This also addresses the portability of this test to allow it to run on OSX since it now builds the servo library during the regular build process for the current machine.

I apologize I did not make this two pull requests since, while I was at it, I also decided to address #274 by removing the hardcoded `lib` `CMAKE_FIND_LIBRARY_PREFIXES` and `.a` `CMAKE_FIND_LIBRARY_SUFFIXES`. This required me to put the TPL package finding after the call to `project(Nalu)` so that the system specific prefixes and suffixes would already be set. I DID NOT do this for Trilinos however (which was a mistake I made last time I attempted this cleanup), as Nalu requires the compilers to be set to Trilinos' before the call to `project(Nalu)`. Then I tried to make the finding of the TPL packages more consistent by using the files CMake installs with each package to use `find_package` which means for YAML we are using `find_package` and not `find_library` as the others do. I also took the liberty of trying to make the CMakeLists more consistent in style and chose lowercasing the commands and making the general formatting more consistent.

I have tested the OpenFAST test on both Linux (Peregrine), and OSX (my laptop), and the test works as I expect and runs successfully. Regarding the searches for TPLs, I have tested this on both Linux (Peregrine), and OSX (my laptop) using both shared and static libraries for OpenFAST, TIOGA, and HYPRE, leaving Trilinos static. On OSX I found that I couldn't built OpenFAST as a shared library and HYPRE does not appear to be able to build as shared on OSX as well. I was able to build YAML and TIOGA as shared, and YAML works fine, but TIOGA may not be adding the full path for itself during linking in Nalu which I have discussed with @sayerhs already, but I feel this does not impact this pull request as I have manually edited the path to the TIOGA shared library in `naluX` and tested successfully. On Linux, everything seemed to work as expected regardless of whether I built the TPLs as shared or static and Nalu was able to find, link and run the tests with them.

I think I would like some assurance from at least one reviewer that this at least does not break the status quo for them by testing they can still build against static libraries most importantly.


CMakeLists.txt
```diff
@@ -12,144 +14,145 @@ option(ENABLE_PARAVIEW_CATALYST
...
+message("   Trilinos_BUILD_SHARED_LIBS = ${Trilinos_BUILD_SHARED_LIBS}")
+message("   Trilinos_CXX_COMPILER_FLAGS = ${Trilinos_CXX_COMPILER_FLAGS}")
+message("End of Trilinos details\n")
+
+message("Setting and checking of compilers:")
+set(CMAKE_CXX_COMPILER ${Trilinos_CXX_COMPILER} )
+set(CMAKE_C_COMPILER ${Trilinos_C_COMPILER} )
+set(CMAKE_Fortran_COMPILER ${Trilinos_Fortran_COMPILER} )
+# Enable Fortran if it is enabled in Trilinos
+if(CMAKE_Fortran_COMPILER)
```
sayerhs: Should this be moved to `project(Nalu CXX Fortran)` ? MasterElements still require the fortran file, so we would require Fortran compiler even if Trilinos is not enabling it? And with `ENABLE_OPENFAST` we definitely need Fortran for `DISCON.f90`

sayerhs:
Combined @jrood-nrel 's branch and it compiles and runs with static libraries for `TIOGA` and `HYPRE`.

CMakeLists.txt
jrood-nrel: Ah, right, I didn't think of that. We probably should move to `project(Nalu CXX Fortran)`. I will do that and do a quick test on it.

PR #304, opened 2018-01-19
Some cleanup in ScratchViews and AssembleElemSolverAlgorithm.

alanw0:
Removed topo argument from ScratchViews and related functions,
and from AssembleElemSolverAlgorithm. Also generalized
entity-rank and nodes-per-entity in AssembleElemSolverAlgorithm.

There's still more to do, but this is a good start.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -89,7 +89,7 @@ public:
 
   ScratchViews(const TeamHandleType& team,
                const stk::mesh::BulkData& bulkData,
-               stk::topology topo,
+               int nodesPerEntity,
                ElemDataRequests& dataNeeded);
 
   virtual ~ScratchViews() {
@@ -412,7 +412,7 @@ void MasterElementViews<T>::fill_master_element_views_new_me(
 template<typename T>
 ScratchViews<T>::ScratchViews(const TeamHandleType& team,
              const stk::mesh::BulkData& bulkData,
-             stk::topology topo,
+             int nodesPerEntity,
              ElemDataRequests& dataNeeded)
 {
   /* master elements are allowed to be null if they are not required */
@@ -421,14 +421,13 @@ ScratchViews<T>::ScratchViews(const TeamHandleType& team,
   MasterElement *meFEM = dataNeeded.get_fem_volume_me();
 
   int nDim = bulkData.mesh_meta_data().spatial_dimension();
-  int nodesPerElem = topo.num_nodes();
   int numScsIp = meSCS != nullptr ? meSCS->numIntPoints_ : 0;
   int numScvIp = meSCV != nullptr ? meSCV->numIntPoints_ : 0;
   int numFemIp = meFEM != nullptr ? meFEM->numIntPoints_ : 0;
 
-  create_needed_field_views(team, dataNeeded, bulkData, nodesPerElem);
+  create_needed_field_views(team, dataNeeded, bulkData, nodesPerEntity);
 
-  create_needed_master_element_views(team, dataNeeded, nDim, nodesPerElem, numScsIp, numScvIp, numFemIp);
+  create_needed_master_element_views(team, dataNeeded, nDim, nodesPerEntity, numScsIp, numScvIp, numFemIp);
 }
 
 template<typename T>
@@ -507,15 +506,12 @@ int get_num_scalars_pre_req_data( ElemDataRequests& dataNeededBySuppAlgs, int nD
 
 void fill_pre_req_data(ElemDataRequests& dataNeeded,
                        const stk::mesh::BulkData& bulkData,
-                       stk::topology topo,
                        stk::mesh::Entity elem,
                        ScratchViews<double>& prereqData,
                        bool fillMEViews = true);
 
 void fill_master_element_views(ElemDataRequests& dataNeeded,
                                const stk::mesh::BulkData& bulkData,
-                               stk::topology topo,
-                               stk::mesh::Entity elem,
                                ScratchViews<DoubleType>& prereqData);
 
 template<typename T = double>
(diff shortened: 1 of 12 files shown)


src/AssembleElemSolverAlgorithm.C
```diff
@@ -85,103 +161,65 @@ AssembleElemSolverAlgorithm::execute()
...
+        extract_vector_lane(smdata.simdrhs, simdElemIndex, smdata.rhs);
+        extract_vector_lane(smdata.simdlhs, simdElemIndex, smdata.lhs);
+        apply_coeff(nodesPerEntity_, entityNodes[simdElemIndex],
+                    smdata.scratchIds, smdata.sortPermutation, smdata.rhs, smdata.lhs, __FILE__);
       }
     });
 
     for(int simdIndex=0; simdIndex<simdLen; ++simdIndex) {
-       delete prereqData[simdIndex];
+       delete smdata.prereqData[simdIndex];
```
sayerhs: Why not use `std::unique_ptr` for this data structure?

src/AssembleElemSolverAlgorithm.C
```diff
@@ -66,6 +68,80 @@ AssembleElemSolverAlgorithm::initialize_connectivity()
...
+{
+    size_t simdBucketLen = bktLength/simdLen;
+    const size_t remainder = bktLength%simdLen;
+    if (remainder > 0) {
+      simdBucketLen += 1;
+    }
+    return simdBucketLen;
+}
+
+struct SharedMemData {
```
sayerhs: Is the thinking that ScratchViews will only be used in ElemSolverAlgorithm? If not, should this be promoted to a header file?

unit_tests/UnitTestKokkosME.h
```diff
@@ -133,17 +133,19 @@ class KokkosMEViews
...
+            copy_and_interleave(prereqData.data(), simdElems, simdPrereqData,
                                 alsoProcessMEViews);
-            fill_master_element_views(dataNeeded_, bulk_, AlgTraits::topo_,
-                                      element, simdPrereqData);
+            fill_master_element_views(dataNeeded_, bulk_, simdPrereqData);
 
             func(simdPrereqData, meSCS_, meSCV_);
           });
+
+          for(int simdIndex=0; simdIndex<simdLen; ++simdIndex) {
```
sayerhs: `std::unique_ptr` instead?

include/ScratchViews.h
```diff
@@ -412,23 +412,22 @@ void MasterElementViews<T>::fill_master_element_views_new_me(
 template<typename T>
 ScratchViews<T>::ScratchViews(const TeamHandleType& team,
              const stk::mesh::BulkData& bulkData,
-             stk::topology topo,
-             ElemDataRequests& dataNeeded)
+             int nodesPerEntity,
```
sayerhs: @alanw0 I didn't understand the motivation to switch from topo to nodesPerEntity, can you explain?

unit_tests/UnitTestKokkosME.h
alanw0: Shreyas, thanks for the detailed review!
Regarding std::unique_ptr, you're right. I'll use that here.

Regarding putting the 'SharedMemData' struct in a header, I'm not sure whether it will be useful outside of this file. I was just trying to clean up the noise inside the bucket loop. But if we put the loop in a function that takes a lambda (like Robert did in the unit-test-kokkos-me stuff), then perhaps this SharedMemData is something we would pass to the lambda... and in that case it belongs in a header.

Regarding the switch from topo to rank and nodes-per-entity, Stefan and I discussed that on Friday and I may not remember all the details, but it seemed cleaner to only pass in the info that's needed. We ultimately are thinking about different types of algorithms, including face and face/elem algorithms. So we don't want ScratchViews for elem-node gathers to have the face topo, and vice-versa.

spdomin: Right.. We want to minimize "topo" since we may have two floating around for a face/element and face combination.

Is this ready to go? It looks good to me, however, I want to make sure @sayerhs feels good about this push. In general, it is only a prep for things to come. As we learn more, some of these things above may change.

sayerhs: @spdomin I am good with this being merged. What is unclear to me is that `ScratchViews` constructor is taking the `nodesPerEntity` instead of `topo`. So is the anticipation that ScratchViews could have multiple `nodesPerEntity` passed to constructor? If not, where is the second `nodesPerEntity` being passed to ScratchViews? My thinking was that Alan had proposed elemViews, faceViews as different ScratchViews instances.

alanw0: That's right, my thinking was that elemViews and faceViews would be separate ScratchViews instances. That could change as we move forward... If they are not separate ScratchViews instances, then we'll have to figure out new arguments to pass in.

sayerhs: @alanw0 thanks for confirming. And if `elemViews` and `faceViews` were different instances then they would only have `topo` associated with it, correct? Sorry, for all these questions. I am just getting up to speed with the design that you and Stefan are working on.

alanw0: The more questions the better. I think we want AssembleElemSolver to not have the topo, since it could be managing both face and elem... Since ScratchViews get constructed within AssembleElemSolver, then they can't have topo either. Thus elemViews would get nodesPerElem, and faceViews would get nodesPerFace. 

This is basically thinking out loud, so nothing is in concrete yet.

spdomin: Yes, that is the point. A topo coming into the ElemAlg constructor was only used for ScratchData. However, if you looked at that code, it was only using it for nodesPerElement.

PR #305, opened 2018-01-22
Fix bad preprocessor directive warnings for DISCON

sayerhs:
Update DISCON file to quell warnings on Intel/Clang compilers.


jrood-nrel:
I guess we could just remove the DLL stuff since we won't be using Windows.

PR #306, opened 2018-01-22
WIP: Run algorithm template

alanw0:
Experiment with templated run_algorithm method.
    
    It is called from within AssembleElemSolverAlgorithm::execute
    and also from the execute method in UnitTestKokkosME.h.
    Borrows heavily from what Robert had done.
    
    Doesn't yet tackle the question of whether a single algorithm
    can run an element-only scheme as well as a face-only or face/elem
    scheme.

### DIFF
--- a/include/AssembleElemSolverAlgorithm.h
+++ b/include/AssembleElemSolverAlgorithm.h
@@ -9,8 +9,13 @@
 #ifndef AssembleElemSolverAlgorithm_h
 #define AssembleElemSolverAlgorithm_h
 
+#include<Realm.h>
 #include<SolverAlgorithm.h>
 #include<ElemDataRequests.h>
+#include <KokkosInterface.h>
+#include<ScratchViews.h>
+#include <SharedMemData.h>
+#include<CopyAndInterleave.h>
 #include<FieldTypeDef.h>
 
 namespace stk {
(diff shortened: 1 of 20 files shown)


include/AssembleElemSolverAlgorithm.h
```diff
@@ -40,6 +54,60 @@ class AssembleElemSolverAlgorithm : public SolverAlgorithm
...
+                           *smdata.prereqData[simdElemIndex], interleaveMEViews_);
+       }
+ 
+       copy_and_interleave(smdata.prereqData, numSimdElems, smdata.simdPrereqData, interleaveMEViews_);
+ 
+       if (!interleaveMEViews_) {
+         fill_master_element_views(dataNeededByKernels_, bulk_data, smdata.simdPrereqData);
+       }
+
+       lambdaFunc(smdata, nodesPerEntity_, entityNodes);
```
spdomin: Let's say that we have a element-based scheme (all hex-mesh) activated with an 1) interior contribution, a 2) wall function, and a 3) symmetry/open. 

Is the concept/design point that we would have three activated ElemAlgs with each being 1) as an interior, 2) a face bc, and 3) two-kernels of a face/element?

If we add a hybrid mesh - well - let's not go there yet:)

alanw0: I'm not sure. I was just reading through AssembleMomentumElemOpenSolverAlg again, trying to think about these kind of questions. I think the 'run_algorithm' method could be the basis for an elem alg or a face alg, with little or no changes. But perhaps a face/elem alg needs a separate 'run_face_elem_algorithm' which would loop over face buckets, get the elem for each face, and manage faceScratchViews and elemScratchViews.

overfelt:
A detailed review/explanation would be very useful for me.  Maybe another afternoon in depth discussion with the team?

spdomin:
Yes, let's find the time to talk through the templating path forward.

alanw0:
I'm going to close this pull request. I updated my branch from nalu:master, and I'm not sure if I did it right. A new pull request can be opened later, after we've made more progress with algorithm design.

PR #307, opened 2018-01-24
Unit-test to prototype a face-element algorithm.

alanw0:
In unit_tests/kernels/UnitTestFaceElemBasic.C see the
class TestFaceElemAlgorithm, and its method run_face_elem_algorithm.

The run_face_elem_algorithm method uses a similar looping structure
to that used in AssembleElemSolverAlgorithm, but manages two
instances of the ElemDataRequests and ScratchViews objects, specifically
for face data and element data.

Jointly with Stefan and Robert, plumbed another master-element ('meFC')
into the ElemDataRequest and ScratchViews objects. We debated whether to
add the new master-element into those objects or to split those objects
such that a given instance would only hold the needed master-elements.
For now at least, we just added a new master-element into the existing
classes.

Still some issues to be worked out:

- Would face-elem-algorithm instances be templated on two AlgTraits, or
on AlgTraits that hold pairs of topologies (for quad-hex, tri-tet, etc).

- the loop in face-elem-algorithm needs nodesPerElem, which is currently
hard-coded for the unit-test but needs to be obtained from stk-mesh or
from a master-element.

- We think AssembleElemSolverAlgorithm could be made general enough to
handle either element (interior) or face (surface) algorithms. But we
haven't specifically designed a class hierarchy or determined where
a face-elem-algorithm would fit into that hierarchy.

- Haven't yet determined exactly which arguments should be passed to
kernel.execute for a face-elem kernel. The prototype in this unit-test
takes some arguments that probably wouldn't be needed in general.

### DIFF
--- a/include/ScratchViews.h
+++ b/include/ScratchViews.h
@@ -48,31 +48,39 @@ public:
   int create_master_element_views(
     const TeamHandleType& team,
     const std::set<ELEM_DATA_NEEDED>& dataEnums,
-    int nDim, int nodesPerElem,
-    int numScsIp, int numScvIp, int numFemIp);
+    int nDim, int nodesPerFace, int nodesPerElem,
+    int numFaceIp, int numScsIp, int numScvIp, int numFemIp);
 
   void fill_master_element_views(
     const std::set<ELEM_DATA_NEEDED>& dataEnums,
     SharedMemView<double**>* coordsView,
+    MasterElement* meFC,
     MasterElement* meSCS,
     MasterElement* meSCV,
-    MasterElement* meFEM);
+    MasterElement* meFEM,
+    int faceOrdinal = -1);
 
   void fill_master_element_views_new_me(
     const std::set<ELEM_DATA_NEEDED>& dataEnums,
     SharedMemView<DoubleType**>* coordsView,
+    MasterElement* meFC,
     MasterElement* meSCS,
     MasterElement* meSCV,
-    MasterElement* meFEM);
+    MasterElement* meFEM,
+    int faceOrdinal = -1);
 
+  SharedMemView<T**> fc_areav;
   SharedMemView<T**> scs_areav;
+  SharedMemView<T***> dndx_fc_scs;
   SharedMemView<T***> dndx;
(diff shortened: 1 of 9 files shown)


unit_tests/kernels/UnitTestFaceElemBasic.C
```diff
@@ -0,0 +1,229 @@
...
+
+int
+calculate_shared_mem_bytes_per_thread(int lhsSize, int rhsSize, int scratchIdsSize, int nDim,
+                                      sierra::nalu::ElemDataRequests& faceDataNeeded,
+                                      sierra::nalu::ElemDataRequests& elemDataNeeded)
+{
+    int bytes_per_thread = (rhsSize + lhsSize)*sizeof(double) + (2*scratchIdsSize)*sizeof(int) +
+                           sierra::nalu::get_num_bytes_pre_req_data<double>(faceDataNeeded, nDim)
+                           +
+                           (rhsSize + lhsSize)*sizeof(double) + (2*scratchIdsSize)*sizeof(int) +
```
alanw0: I just noticed this bytes_per_thread calculation is wrong... I'll add a commit which fixes it.

include/ScratchViews.h
```diff
@@ -571,7 +572,7 @@ void fill_pre_req_data(ElemDataRequests& dataNeeded,
 void fill_master_element_views(ElemDataRequests& dataNeeded,
```
spdomin: I still like the possible path forward to use unique names since it is the AssembleElemAlg that calls this method. Moreover, in the arg, it "knows" whether a face or element data is required. This can be a possible goal. In the meantime, this looks like a good step forward.

alanw0: I like the unique names path too. perhaps we could get together later and determine the lists of master-element calls that would belong in each function if we were to split things into unique names.

spdomin: Will do.. In the meantime, feel free to process the push. If you still do not have permissions, let me know and I can do it. @rcknaus is making progress on the exposed_grad_op and the plumbing. I think that I can convert a wall function algorithm to use the new meFC interface in not much time.

PR #310, opened 2018-01-30
Design order Ksgs

spdomin:


spdomin:
Thanks @alanw0. We have many "global" constants that are fixed over the simulation.

I will add a unit test today or tomorrow in addition to a simulation case.

sayerhs:
@spdomin If they are constant for the duration of the lifetime of the kernel, why not declare them as `const` member variables?

overfelt:
So, to show intent, could make cEps_ and tkeProdLimitRation_ const?

alanw0:
@sayerhs yes I think they could be const since they are set in the constructor initializer list. That would be a nice way to have the code document the fact that they won't be written to.

sayerhs:
I would also recommend removing the default initializer from the header files.

spdomin:
I think that I am half way between two competing use cases. In general, such constants are - well - constant. In this case, they should be const. In our full use cases, sometimes these constants can change over time in a simulation by user "steering". However, if that is the case, I am missing the setup to re-set the values. I will move forward with a pure const approach and see if the dynamic steering is a use case about which we care.

PR #313, opened 2018-02-02
New face/element symmetry kernel with unit tests (pass is that it runs cleanly)

spdomin:
* Many changes to the Scratch/Elem interface to drive correctness
in the meFC and meFC/meSCS/meSCV combination with the basic meSCS
kernel design.

* adding more error checks for bad face/element data on respective
face or element data.

* added edge and face rank for ip fields.

* still sharing face and element through Scratch functions. Seems okay
  however, now we have an ScratchMeInfo.

* Extra Ksgs change to const member data.

Clean tests!!

Notes:

a) need an interface to meFC->shape_function
b) need a coded new me kokkos face_grad_op and area vector. Never
   support non-kokkos calls though the new Kernel interface.
c) face ordinals needs to be provided to fill_me.
d) consolidate BcAlgTraits to something more scalable, although
at first, only concentrate on quad4/hex8 pairs.
e) lots of places where nodesPerElement_, numFaceIps_, etc., is specified
   - need to fix this.


spdomin:
Paired with Alan and Robert. Lots of WIP here, however, the design looks clean for a first shot. This push will allow Alan to create the specialized FaceElemSolverAlg class. Then, interior and face albs will be covered by the current ElemSolverAlg while the FaceElemSolverAlg to manage complex bcs. Robert is working on the face_grad_op which is a bit complex by the fact that we need a SIMD interface on a switch for face grad op for the hybrid elements, e.g., pyramid and wedge, e.g., tri/wedge or quad/wedge.

include/BcAlgTraits.h
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef BcAlgTraits_h
+#define BcAlgTraits_h
+
+#include <stk_topology/topology.hpp>
+
+namespace sierra {
+namespace nalu {
+
+struct BcAlgTraitsHex8Quad4 {
```
sayerhs: Is the plan to write out every possible combination in this header file? Or is this a placeholder that will be eventually replaced by some templates that are derived from `struct`s in `AlgTraits.h`?

include/ScratchViews.h
```diff
@@ -484,16 +503,27 @@ ScratchViews<T>::ScratchViews(const TeamHandleType& team,
...
   int numFemIp = meFEM != nullptr ? meFEM->numIntPoints_ : 0;
 
-  create_needed_field_views(team, dataNeeded, bulkData, nodesPerEntity);
+  create_needed_field_views(team, dataNeeded, bulkData, nodalGatherSize);
 
   create_needed_master_element_views(team, dataNeeded, nDim, nodesPerFace, nodesPerElem, numFaceIp, numScsIp, numScvIp, numFemIp);
 }
 
+template<typename T>
+ScratchViews<T>::ScratchViews(const TeamHandleType& team,
```
sayerhs: What is the motivation for a second constructor here? It is only used for elemViews?

include/BcAlgTraits.h
spdomin: This is a placeholder. The plan is to add a templated method that instantiates each unique pair based on AlgTrait combinations.

include/ScratchViews.h
spdomin: The complexity here is that for a face or interior algorithm, we have all we need, e.g., meFC or meSCS. Moreover, the nodesPerEntity is uniquely defined by either nodesPerFace or nodesPerElement. For the face/element combination, the developer needs are more complex. For example, we might need both meFC and meSCS. Worse yet, an exposed face grad op requires meSCS->face_grad_op(). However, this is sized by numFacIp*nodesPerElem*nDim. However, meSCS has not interface to providing numFaceIp for each ordinal. The best path forward, at least for now, seemed to be to create a new ScratchViews constructor with all of the information packed up in a struct.

sayerhs: @spdomin Thanks for the explanation.

spdomin: There is a lot of nitty gritty here. Let's meet as a team next week and talk through some of the current complexities and how to move forward. At present, we are managing face/elem, face, and elem data through one ElemDataRequest type. We seem to go back and forth with considering breaking this into two data requests types. Also, nodal gathering is somewhat independent of master element calls with exception of shared nodal fields - solely "coordinates".

The plan is to start iterating on how to make things more clean. In fact, it may be fine now, we just need more experience with writing face and face/elem kernels along with the accompanied FaceElemSolverAlg (@alanw0, you can start on that this morning and also process this pull request if it looks consistent with what we were doing the other day).

At the very least, this allows us to move forward with all of the other required pieces to start a full face/elem kernel approach. My plan is to deprecate the symmetry bc ASAP.

alanw0: @sayerhs Regarding the second constructor for ScratchViews: we also discussed the possibility of creating a DataRequests base-class, with ElemDataRequests and FaceDataRequests derived classes. That would still require an overloaded constructor etc for ScratchViews. We can keep iterating on the best design as we gain more experience with face-element algorithms.

@spdomin I'll work on elevating the faceElemAlg class from the unit-test to a "real" AssembleFaceElemSolverAlgorithm class.

PR #314, opened 2018-02-02
Promote unit-test face-elem-alg class to

alanw0:
AssembleFaceElemSolverAlgorithm.

### DIFF
--- a/include/AssembleFaceElemSolverAlgorithm.h
+++ b/include/AssembleFaceElemSolverAlgorithm.h
@@ -1,0 +1,124 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+
+#ifndef AssembleFaceElemSolverAlgorithm_h
+#define AssembleFaceElemSolverAlgorithm_h
+
+#include <SolverAlgorithm.h>
+#include <ElemDataRequests.h>
+#include <ScratchViews.h>
+#include <SimdInterface.h>
+#include <SharedMemData.h>
+
+namespace stk {
+namespace mesh {
+class Part;
+}
+}
+
+namespace sierra{
+namespace nalu{
+
+class Realm;
+
+class AssembleFaceElemSolverAlgorithm : public SolverAlgorithm
+{
+public:
+  AssembleFaceElemSolverAlgorithm(
+    Realm &realm,
+    stk::mesh::Part *part,
+    EquationSystem *eqSystem,
+    unsigned nodesPerFace,
+    unsigned nodesPerElem,
+    bool interleaveMeViews = true);
+  virtual ~AssembleFaceElemSolverAlgorithm() {}
+  virtual void initialize_connectivity();
+  virtual void execute();
+
+  template<typename LambdaFunction>
+  void run_face_elem_algorithm(stk::mesh::BulkData& bulk, LambdaFunction lamdbaFunc)
+  {
+      int nDim = bulk.mesh_meta_data().spatial_dimension();
+
+      sierra::nalu::MasterElement* meFC = faceDataNeeded_.get_cvfem_face_me();
+      sierra::nalu::MasterElement* meSCS = faceDataNeeded_.get_cvfem_surface_me();
+      sierra::nalu::MasterElement* meSCV = faceDataNeeded_.get_cvfem_volume_me();
+
+      sierra::nalu::ScratchMeInfo meElemInfo;
+      meElemInfo.nodalGatherSize_ = nodesPerElem_;
+      meElemInfo.nodesPerFace_ = nodesPerFace_;
+      meElemInfo.nodesPerElement_ = nodesPerElem_;
+      meElemInfo.numFaceIp_ = meFC != nullptr ? meFC->numIntPoints_ : 0;
+      meElemInfo.numScsIp_ = meSCS != nullptr ? meSCS->numIntPoints_ : 0;
+      meElemInfo.numScvIp_ = meSCV != nullptr ? meSCV->numIntPoints_ : 0;
+
+      int rhsSize = meElemInfo.nodalGatherSize_*numDof_, lhsSize = rhsSize*rhsSize, scratchIdsSize = rhsSize;
+
+      const int bytes_per_team = 0;
+      const int bytes_per_thread = calculate_shared_mem_bytes_per_thread(lhsSize, rhsSize, scratchIdsSize,
+                                                                       nDim, faceDataNeeded_, elemDataNeeded_, meElemInfo);
+
+      const bool interleaveMeViews = false;
+
+      stk::mesh::Selector s_locally_owned_union = bulk.mesh_meta_data().locally_owned_part() & *part_;
+      stk::mesh::EntityRank sideRank = bulk.mesh_meta_data().side_rank();
+      stk::mesh::BucketVector const& buckets = bulk.get_buckets(sideRank, s_locally_owned_union );
+
+      auto team_exec = sierra::nalu::get_team_policy(buckets.size(), bytes_per_team, bytes_per_thread);
+      Kokkos::parallel_for(team_exec, [&](const sierra::nalu::TeamHandleType& team)
+      {
+        stk::mesh::Bucket & b = *buckets[team.league_rank()];
+
+        ThrowAssertMsg(b.topology().num_nodes() == (unsigned)nodesPerFace_,
+                       "TestFaceElemAlgorithm expected nodesPerEntity_ = "
+                       <<nodesPerFace_<<", but b.topology().num_nodes() = "<<b.topology().num_nodes());
+
+        SharedMemData_FaceElem smdata(team, bulk, faceDataNeeded_, elemDataNeeded_, meElemInfo, rhsSize);
+
+        const size_t bucketLen   = b.size();
+        const size_t simdBucketLen = sierra::nalu::get_num_simd_groups(bucketLen);
+
+        Kokkos::parallel_for(Kokkos::TeamThreadRange(team, simdBucketLen), [&](const size_t& bktIndex)
+        {
+          smdata.numSimdFaces = sierra::nalu::get_length_of_next_simd_group(bktIndex, bucketLen);
+
+          for(int simdFaceIndex=0; simdFaceIndex<smdata.numSimdFaces; ++simdFaceIndex) {
+            stk::mesh::Entity face = b[bktIndex*simdLen + simdFaceIndex];
+            smdata.elemFaceOrdinals[simdFaceIndex] = bulk.begin_element_ordinals(face)[0];
+            sierra::nalu::fill_pre_req_data(faceDataNeeded_, bulk, face, *smdata.faceViews[simdFaceIndex], interleaveMeViews);
+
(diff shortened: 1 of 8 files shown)


include/Kernel.h
```diff
@@ -91,9 +91,19 @@ class Kernel
    *  the linear solve
    */
   virtual void execute(
-    SharedMemView<DoubleType**>&,
-    SharedMemView<DoubleType*>&,
-    ScratchViews<DoubleType>&)
+    SharedMemView<DoubleType**> &lhs,
```
sayerhs: Wouldn't this create unused variable warnings?

src/AssembleElemSolverAlgorithm.C
```diff
@@ -68,49 +69,6 @@ AssembleElemSolverAlgorithm::initialize_connectivity()
...
-{
-    size_t simdBucketLen = bktLength/simdLen;
-    const size_t remainder = bktLength%simdLen;
-    if (remainder > 0) {
-      simdBucketLen += 1;
-    }
-    return simdBucketLen;
-}
-
 struct SharedMemData {
```
sayerhs: Consolidate this in `SharedMemData.h`?

include/Kernel.h
alanw0: I wouldn't think that should create unused variable warnings. If it does, I guess we could either comment out the argument, e.g., /*lhs*/ or remove the curly braces to make the method pure virtual. In general I find it annoying to see prototypes with only types but no argument names.

src/AssembleElemSolverAlgorithm.C
alanw0: Yes. I will put that into SharedMemData.h.

include/Kernel.h
spdomin: I thought that the header can have all of the variable names while if an implementation does not need the variable, one would:

SharedMemView<DoubleType**> &/*lhs*/

I do not prefer the bare signatures, however, am fine with whomever is writing the class/kernel to decide:)

sayerhs: Yeah, Alan had provided empty implementations in the header (so method definition and not declaration), so I thought it would issue warnings. However, it seems that Clang on OSX  doesn't complain about it. Sorry for the noise.

src/AssembleElemSolverAlgorithm.C
sayerhs: I was expecting this: 

```
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/src/ContinuityAdvElemKernel.C:8:
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/include/ContinuityAdvElemKernel.h:11:
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:94:34: warning:
      unused parameter 'lhs' [-Wunused-parameter]
    SharedMemView<DoubleType**> &lhs,
                                 ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:95:33: warning:
      unused parameter 'rhs' [-Wunused-parameter]
    SharedMemView<DoubleType*> &rhs,
                                ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:96:31: warning:
      unused parameter 'scratchViews' [-Wunused-parameter]
    ScratchViews<DoubleType> &scratchViews)
                              ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:103:34: warning:
      unused parameter 'lhs' [-Wunused-parameter]
    SharedMemView<DoubleType**> &lhs,
                                 ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:104:33: warning:
      unused parameter 'rhs' [-Wunused-parameter]
    SharedMemView<DoubleType*> &rhs,
                                ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:105:31: warning:
      unused parameter 'faceScratchViews' [-Wunused-parameter]
    ScratchViews<DoubleType> &faceScratchViews,
                              ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kernel.h:106:31: warning:
      unused parameter 'elemScratchViews' [-Wunused-parameter]
    ScratchViews<DoubleType> &elemScratchViews)
                              ^
235 warnings generated.
```

include/AssembleFaceElemSolverAlgorithm.h
```diff
@@ -0,0 +1,124 @@
...
+
+namespace sierra{
+namespace nalu{
+
+class Realm;
+
+class AssembleFaceElemSolverAlgorithm : public SolverAlgorithm
+{
+public:
+  AssembleFaceElemSolverAlgorithm(
```
spdomin: This class will be computed based on the known face/element pair - which will be the same for all kernels attached to it. It may be best to expect that this class provides the scratchMeStruct behind the scenes.

alanw0:
Regarding the style of having a 'run_algorithm' template which takes a lambda:
You're right it is similar to the other pull request. I had done the other one so that the same code could be used by unit-tests (such as the tests in UnitTestKokkosME) rather than having to duplicate the kokkos loops just for the unit-test.
In this case the same thing is happening in the faceElemBasic test versus your faceElemMomentumSymmetry test: i.e., the basic test doesn't compute linear-system contributions, but can still make use of the run_algorithm code rather than duplicating it.
Whereas your faceElemMomentumSymmetry test is better suited to calling the execute method which goes ahead and manages linear-system contributions.

src/AssembleElemSolverAlgorithm.C
alanw0: @sayerhs I'm not clear whether you are saying you are seeing these warnings (meaning something needs to be changed or fixed in Kernel.h) or not?

sayerhs:
Hi Alan,

Sorry  I wasn't being clear. With just `-Wall` I don't see the warning. I
need to turn on `-Wextra` as well to see the warning I posted.

spdomin:
@sayerhs, what is the formal rule?

MyClass.C:

```
MyClass::MyClass( int one, int /*two*/) {
  one = 10; // two not used
}
```

would require:

MyClass.h

```
public:
  MyClass (int one, int);

```

As such, for safety, you always leave the header without variable names?

Thanks...

sayerhs:
@spdomin 

In the example you provided, since the definition is in the `.C` file, you can safely have `(int one, int two)` in the header file. I'll use the following example to try and answer your question. 

```
// File test.h
int testfunc(int a, int b);

int testfunc2(int a, int b)
{
    return 0;
}

// File test.cpp
#include "test.h"

int testfunc(int a, int b)
{
    return a+b;
}

int main()
{
    return 0;
}

$ c++ -Wall -Wextra -pedantic test.cpp
In file included from test.cpp:1:
./test.h:4:19: warning: unused parameter 'a' [-Wunused-parameter]
int testfunc2(int a, int b)
                  ^
./test.h:4:26: warning: unused parameter 'b' [-Wunused-parameter]
int testfunc2(int a, int b)
                         ^
./test.h:5:2: warning: control reaches end of non-void function [-Wreturn-type]
{}
 ^
3 warnings generated.
(sci) sanantha@gandiva:~/Downloads$ c++ -Wall -Wextra -pedantic test.cpp
In file included from test.cpp:1:
./test.h:4:19: warning: unused parameter 'a' [-Wunused-parameter]
int testfunc2(int a, int b)
                  ^
./test.h:4:26: warning: unused parameter 'b' [-Wunused-parameter]
int testfunc2(int a, int b)
                         ^
2 warnings generated.
```

You can have variables in _function declarations_ in header files and the compilers won't generate the `unused-parameter` warning (e.g., `testfunc` in my example). This makes sense because at this point the compiler hasn't yet seen the definition to make the determination as to whether the parameters are, in fact, unused. This is not the case for `testfunc2` where it does see the definition and raises the warning (no different than when the definition occurs in a `.cpp` file). 

I agree with Alan that having variable names are nice in the headers, particularly when we use doxgen-capable comments for documentation in headers. For the `Kernel.h` functions since we are providing default implementations they will have to be commented out to prevent warnings with `-Wextra`

```
  /** Execute the kernel within a Kokkos loop and populate the LHS and RHS for
   *  the linear solve
   */
  virtual void execute(
    SharedMemView<DoubleType**>& /* lhs */,
    SharedMemView<DoubleType*>& /* rhs */,
    ScratchViews<DoubleType>& /* scratchViews */)
  {}
```

I drop them because I feel I am not providing any useful information to the compiler. However, from a readability point of view this is better than not providing the variable names as someone will have to look at the C file to figure out what these variables are. On the flip side, the compiler will not complain if the header declaration had been 

```
  virtual void execute(
    SharedMemView<DoubleType**>&  lhs ,
    SharedMemView<DoubleType*>& scratchViews,
    ScratchViews<DoubleType>& rhs);
```

sayerhs:
@spdomin I realized that the example in my previous post did not address your question fully. Here is an updated example

```
///////////////////////////////
// test.h

int testfunc(int a, int b);

int testfunc2(int a, int b)
{
    return 0;
}

//////////////////////////////
// test.cpp
#include "test.h"

int testfunc(int a, int)
{
    return 2*a;
}

int main()
{
    return 0;
}

$ c++ -Wall -Wextra -pedantic test.cpp
In file included from test.cpp:1:
./test.h:4:19: warning: unused parameter 'a' [-Wunused-parameter]
int testfunc2(int a, int b)
                  ^
./test.h:4:26: warning: unused parameter 'b' [-Wunused-parameter]
int testfunc2(int a, int b)
                         ^
2 warnings generated.
```

spdomin:
Thank you @sayerhs for these examples. I still wonder why many of our virtual functions in the master element class, which throw if called, are not providing warnings. Or are they?

sayerhs:
@spdomin Those generate warnings too, if you have `-Wextra` flag enabled. 

```
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/src/Continui
tyAdvElemKernel.C:8:
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Cont
inuityAdvElemKernel.h:11:
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Kern
el.h:13:
In file included from /Users/sanantha/Documents/nrel/code/nalu/Nalu/include/Scra
tchViews.h:18:
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/master_element/MasterEleme
nt.h:55:34: warning: unused parameter 'shpfc' [-Wunused-parameter]
    SharedMemView<DoubleType**> &shpfc) {
                                 ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/master_element/MasterEleme
nt.h:59:34: warning: unused parameter 'shpfc' [-Wunused-parameter]
    SharedMemView<DoubleType**> &shpfc) {
                                 ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/master_element/MasterEleme
nt.h:63:33: warning: unused parameter 'coords' [-Wunused-parameter]
    SharedMemView<DoubleType**>&coords,
                                ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/master_element/MasterElement.h:64:34: warning: unused parameter 'gradop' [-Wunused-parameter]
    SharedMemView<DoubleType***>&gradop,
                                 ^
/Users/sanantha/Documents/nrel/code/nalu/Nalu/include/master_element/MasterElement.h:65:34: warning: unused parameter 'deriv' [-Wunused-parameter]
    SharedMemView<DoubleType***>&deriv) {
```

spdomin:
Yes, now that we all know about this, let's start cleaning them up. Many thanks for teaching us about the compiler warnings.

Issue #312, opened 2018-02-02
Consistent boundary data for wall functions


spdomin:
When using wall functions, we need to ensure that:

1) the projected nodal gradient is using velocity_bc (this should be consistent with the projected nodal gradient equation system owned by momentum)
2) the projection is active - unlike in Dirichlet conditions
3) the velocity_bc is not copied into the dof value in the wall registration method. In this case, we may want to add the algorithm to the IC that type so that we can evaluate if we should populate the data at the boundary or let the IC prevail.

Once these three are complete, I would like to run a few canonical wall function runs.

spdomin:
Here is a patch... I think we need to make each change separate and the see if anything diffs.

[wallBcFix.txt](https://github.com/NaluCFD/Nalu/files/1690159/wallBcFix.txt)

spdomin:
@rcknaus, the projected nodal gradient equation system has a call set_data_map("fieldName"). The velocity management looks similar to this issue here. Speficially, the dof is used for the boundary contributions. The complexity in this system is that we do not solve the Gjui system, rather, GjQ with Q being copied in by the current velocity. I think we need boundary data within this copy...

Thoughts?

spdomin:
the above patch was a bit off...
```

  // wall function activity will only set dof velocity np1 wall value as an IC
  if ( anyWallFunctionActivated )
    realm_.initCondAlg_.push_back(theCopyAlg);
  else
    bcDataMapAlg_.push_back(theCopyAlg);
```

I am also thinking through what user data should be used where, most notably:

`const std::string bcFieldName = wallFunctionApproach ? "wall_velocity_bc" : "velocity_bc";`

The above looks to have missed the ABL approach.

spdomin:
Completed. Will create a new ticket for the PNG for momentum.

PR #317, opened 2018-02-06
Be more careful with wall data in presence of Dirichlet or wall fuction.

spdomin:
Addresses issue: https://github.com/NaluCFD/Nalu/issues/312

* Use specified inflow velocity for Gjui inflow contribution. Formerly,
we were always using the dof velocity. As only Dirichlet inflow bcs are
supported, this will not chnage anything, however, will be appropriate for
"use_fluxes" possible implementation.

* Use specified wall velocity for Gjui wall contribution. Formerly,
we were always using the dof velocity. When wall fucntions are used, we
want the specified velocity from the user input file.

* Do not project only when Dirichlet values are specified in wall bc; any type of wall-function will require projection.

* Do not popuate the dof velocity with user specified value each time step when using wall functions. Rather, only set the IC value to the wall. Again, Dirichlet's win...

* wall_velocity_bc or velocity_bc usage? In the ABL use-case, this was wrong. At any wall function, we use wall_velocity_bc as boundary data for projected nodal gradients and boundary contributions.

* new gold files due to the above changes. All verified to be proper. In general,
most cases look like equal or lower norms. In some cases, norms are higher now.

Note:

a) the movingCylinder was found to diff based on the change of velocityNp1 to bcVelocity.
However, the new code is proper - order-of-ops.

b) the code could benifit from a use_fluxes inflow now. However, that would require
property management improvements.


spdomin:
This is good to go now as all tests have been blessed.

spdomin:
Th aka for the review. Matt B had it correct:) we follow this approach for the VC codes.  Once we have the use-flux inflow option things will be one level cleaner.

PR #319, opened 2018-02-07
Add kernel directory in prep for new bc conversions

spdomin:
before I start on a few face/elem kernels, I wanted a restructure.

@marchdf and @jamelvin, this may provide a few conflicts in your WIP turbulence modeling in kernels. I hope that this is not too disruptive. If it is, let me know. Otherwise, the NGP apps team agreed that a high level restructure is useful..


jamelvin:
@spdomin Shouldn't be a big issue for what we are doing, seems easy enough to modify what we have.  Thanks.

spdomin:
I tested as well:)

PR #320, opened 2018-02-07
WIP - Code does not build Add Momentum open adv/diff kernel

spdomin:
Notes:

a) several complexities shown in all of the fixme statements.

### DIFF
--- a/src/kernel/MomentumOpenAdvDiffElemKernel.C
+++ b/src/kernel/MomentumOpenAdvDiffElemKernel.C
@@ -1,0 +1,370 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#include "kernel/MomentumOpenAdvDiffElemKernel.h"
+#include "BcAlgTraits.h"
+#include "master_element/MasterElement.h"
+#include "SolutionOptions.h"
+
+// template and scratch space
+#include "BuildTemplates.h"
+#include "ScratchViews.h"
+
+// stk_mesh/base/fem
+#include <stk_mesh/base/Entity.hpp>
+#include <stk_mesh/base/MetaData.hpp>
+#include <stk_mesh/base/Field.hpp>
+
+namespace sierra {
+namespace nalu {
+
+template<typename BcAlgTraits>
+MomentumOpenAdvDiffElemKernel<BcAlgTraits>::MomentumOpenAdvDiffElemKernel(
+  const stk::mesh::MetaData &metaData,
+  const SolutionOptions &solnOpts,
+  VectorFieldType *velocity,
+  GenericFieldType *Gjui,
+  ScalarFieldType *viscosity,
+  ElemDataRequests &faceDataPreReqs,
+  ElemDataRequests &elemDataPreReqs)
+  : Kernel(),
+    viscosity_(viscosity),
+    Gjui_(Gjui),
+    alphaUpw_(solnOpts..get_alpha_upw_factor("velocity")),
+    om_alphaUpw_(1.0-alphaUpw_),
+    hoUpwind_(solnOpts.get_upw_factor("velocity")),
+    nfEntrain_(solnOpts.nearestFaceEntrain_),
+    om_nfEntrain_(1.0-nfEntrain_),
+    includeDivU_(solnOpts.includeDivU_),
+    shiftedGradOp_(solnOpts.get_shifted_grad_op(velocity->name())),
+    ipNodeMap_(sierra::nalu::MasterElementRepo::get_surface_master_element(BcAlgTraits::elemTopo_)->ipNodeMap()),
+    faceIpNodeMap_(sierra::nalu::MasterElementRepo::get_face_master_element(BcAlgTraits::faceTopo_)->ipNodeMap()),
+{
+  // save off fields
+  velocityNp1_ = &(velocity->field_of_state(stk::mesh::StateNP1));
+  if ( solnOpts.does_mesh_move() )
+    velocityRTM_ = msetaData.get_field<VectorFieldType>(stk::topology::NODE_RANK, "velocity_rtm");
+  else
+    velocityRTM_ = velocity;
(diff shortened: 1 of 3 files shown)


spdomin:
As noted, this is a WIP and intended to drive discussion on the possible missing pieces to the Face/Element kernel design.

spdomin:
Neglect this fix me....
 const DoubleType densL   = v_density(nodeL);
 const DoubleType densR   = *stk::mesh::field_data(densityNp1, nodeR);

src/kernel/MomentumOpenAdvDiffElemKernel.C
```diff
@@ -0,0 +1,370 @@
...
+    const DoubleType om_pecfac = 1.0-pecfac;
+   
+    //================================
+    // advection first
+    //================================
+    const DoubleType tmdot = vf_openMassFlowRate(ip);
+    
+    // advection; leaving the domain
+    // fixme... oh my, do I need to wrap this in a stk::math::if_then_else(tmdot > 0, doThis, doThat);
+    if ( tmdot > 0.0 ) {
```
alanw0: Yes you either need to use stk::math::if_then_else, or put in a loop over simd::ndoubles as is currently done above for the unconverted pecletFunction.

src/kernel/MomentumOpenAdvDiffElemKernel.C
```diff
@@ -0,0 +1,370 @@
...
+  SharedMemView<DoubleType**>& v_velocityNp1 = elemScratchViews.get_scratch_view_2D(*velocityNp1_);
+  SharedMemView<DoubleType**>& v_vrtm = elemScratchViews.get_scratch_view_2D(*velocityRTM_);
+  SharedMemView<DoubleType**>& v_viscosity = elemScratchViews.get_scratch_view_1D(*viscosity_);
+  SharedMemView<DoubleType**>& v_density = elemScratchViews.get_scratch_view_1D(*density_);
+  SharedMemView<DoubleType***>& v_dndx = shiftedGradOp_
+    ? elemScratchViews.get_me_views(CURRENT_COORDINATES).dndx_shifted_fc_scs
+    : elemScratchViews.get_me_views(CURRENT_COORDINATES).dndx_fc_scs;
+
+  // fixme: need master elements
+  MasterElement* meFC = NULL;
```
spdomin: I would like to avoid something like:

`MasterElement *meSCS = sierra::nalu::MasterElementRepo::get_surface_master_element(BcAlgTraits::elemTopo_);`

src/kernel/MomentumOpenAdvDiffElemKernel.C
```diff
@@ -0,0 +1,370 @@
...
+    const int face_ordinal = -1;
+
+    // fixme: need the master element
+    const int opposingNode = meSCS->opposingNodes(face_ordinal,ip);
+    const int nearestNode = ipNodeMap_[ip];
+    const int opposingScsIp = meSCS->opposingFace(face_ordinal,ip);
+    const int localFaceNode = faceIpNodeMap[ip];
+    
+    // fixme: left and right nodes; right is on the face; left is the opposing node
+    stk::mesh::Entity nodeL = stk::mesh::Entity(); // elem_node_rels[opposingNode];
```
spdomin: The nodeL and nodeR can be removed by simply extracting data off of, e..g, v_phi(nearestNode);

src/kernel/MomentumOpenAdvDiffElemKernel.C
```diff
@@ -0,0 +1,370 @@
...
+        
+        // upwind lhs
+        lhs(indexR, nearestNode*BcAlgTraits::nDim_+i) += tmdot*pecfac*alphaUpw;
+        
+        // central part
+        const DoubleType fac = tmdot*(pecfac*om_alphaUpw+om_pecfac);
+        for ( int ic = 0; ic < NcAlgTraits::nodesPerFace_; ++ic ) {
+          const DoubleType r = vf_shape_function(ip,ic);
+          // fixme: where to extract face_node_ordinals
+          const int nn = face_node_ordinals[ic];
```
spdomin: I think this is still not clear to me....

alanw0: If we need face-node-ordinals here, we may need to consider passing in some struct similar to the SharedMemData struct that is held in run_face_elem_algorithm but which is currently not passed down to kernel->execute. Currently that struct has elem-face-ordinals. It has a place-holder for connectedNodes but that isn't yet populated. We'll have to discuss just how much stuff we want(need) to populate and pass down.

src/kernel/MomentumOpenAdvDiffElemKernel.C
overfelt: Why?  
Save it off during the constructor instead?

alanw0: Yes that's what I was going to suggest. The master-elements are set/available in the kernel constructor, they could be set as members at that point.

spdomin: Right... Done:)

src/kernel/MomentumOpenAdvDiffElemKernel.C
spdomin: I may pick the later as indents can get out of control... I think.. I will try both and see.

src/kernel/MomentumOpenAdvDiffElemKernel.C
spdomin: I think that @rcknaus is dealing with this for his master element exposed face grad_op... Thoughts Robert?

spdomin:
@alanw0, were you joking about the 250 lines? The original class implementation is `550 lines total whereas the new one is ~350. Aside from these small items, I think it looks okay.

alanw0:
@spdomin 350 lines is certainly better than 550! so that's progress. But I was thinking about both testability and readability, and those are both harder, the bigger a function is. I wonder if the execute function could benefit from having some of its code sections extracted into free-standing functions. That would have a couple of benefits: 1. those functions could be unit-tested and 2. the execute method would then be much easier to read.

spdomin:
We may be able to separate advection from diffusion. As for unit testing, I thought that the full Kernel was easily tested?

src/kernel/MomentumOpenAdvDiffElemKernel.C
spdomin: Is there any cost to using this inline function for such a large code fragment? I went with the if_then_else....

spdomin:
Thanks for the discussion. I have a new iteration and will post a new PR once I have the unit test.

PR #321, opened 2018-02-08
Support SIMD and non-SIMD versions of PecletFunction

sayerhs:


sayerhs:
@spdomin @rcknaus I would be interested in hearing your thoughts on this. All the regression tests run. I would like to add unit tests for `PecletFunction` if you guys agree that this design is good. Alternate option would be to create `PecletFunctionSIMD` and use those in the Kernels, but I am not very happy with that design approach.

include/EquationSystem.h
```diff
@@ -298,6 +299,28 @@ class EquationSystem
...
+//-------- create_peclet_function ------------------------------------------
+//--------------------------------------------------------------------------
+template<typename T>
+PecletFunction<T>*
+EquationSystem::create_peclet_function(
+  const std::string dofName)
+{
+  PecletFunction<T> *pecletFunction = NULL;
+  if ( "classic" == realm_.get_tanh_functional_form(dofName) ) {
+    const double hybridFactor = realm_.get_hybrid_factor(dofName);
```
sayerhs: Should probably use `static_cast` here.

alanw0:
I agree it would be nice to include a unit test.
Ideally I advocate leading with the test, i.e., write the test first then write the production code. But I acknowledge that is often difficult when modifying existing code rather than writing brand new code. I've come to believe that the importance of unit-testing can't be overstated.

spdomin:
@sayerhs, many thanks for this fast turn around.

@alanw0, I agree about unit testing. What I am finding is that determination of correctness and/or the gold standards is generally hard in our field. In the case of a Peclet number, probably not. 

We need some sort of foundational effort to start figuring out how to gauge correctness of a complex  kernel. Any ideas?

sayerhs:
@spdomin @alanw0 I just want to point out that I did add unit tests before merging. Please see [b2e3e46](https://github.com/NaluCFD/Nalu/pull/321/commits/b2e3e46d74eb312f6e47be7d4fd7e4389b9cda82).

alanw0:
@spdomin I think that the "unit" in unit-testing probably needs to be smaller than a whole kernel::execute in many cases. If kernel::execute implements several mathematical equations or expressions, perhaps individual expressions (or terms in large expressions) should be functions, and then those could be unit-tested. That way perhaps they would also have simpler inputs and could be unit-tested without having to set up a mesh or an algorithm with realm etc etc. Also multiple kernels could possibly share some of those functions rather than duplicating the code.

spdomin:
@alanw0, I like the sound of this approach and would like to work on this as part of our NGP-apps team. @sayerhs and I have chatted about such topics before. If you (Alan) can take a look at a few of the kernels we have already written to see how to improve things, that would be wonderful.

Thus far, I have been thinking that the smallest "unit test" for kernels is the kernel itself.

PR #322, opened 2018-02-09
Adding system include specifier in CMake to suppress TPL header warnings

jrood-nrel:
This is able to turn off warnings from the TPL header files in a decent manner in my opinion and I have tested it for GCC and Clang. According to the CMake documentation, this is mostly used to suppress header warnings, but might have minor side effects involving treating the header dependencies differently depending on the compiler. For GCC and Clang, their documentation indicates that treating headers as system headers seems to only affect the diagnostics.

@sayerhs What do you think about this solution?


sayerhs:
This is a pretty good solution to allow us to turn on all warnings on Nalu without the noise from Boost/Trilinos/YAML-CPP etc. that we cannot control. This effectively ends up using `-isystem` instead of `-I` that will affect search order, but I don't anticipate issues with this during build because our current use cases don't involve changes to Trilinos headers while developing Nalu. 

I would recommend you add this to the YAML, hypre, and may be TIOGA include lines also. OpenFAST is being developed by members who are also on Nalu team, so it would be nice to see those warnings, @gantech and @michaelasprague what do you think?

jrood-nrel:
Sure. I will wait on TIOGA. I agree about OpenFAST. I will be curious to turn on `-Wextra` after this.

PR #323, opened 2018-02-09
Update YAML-CPP Instructions in Documentation

jrood-nrel:
This updates the recommendation when building YAML-CPP to use a specific commit or its master branch.


jrood-nrel:
I don't like that this isn't showing up as one commit. Let me try this again.

PR #327, opened 2018-02-09
Adding options to enable warnings

jrood-nrel:
Since I've been looking at warnings today, I've just remembered that `-Wall` is not enabled by default, which is what I have the testing set for. To get everyone on the same page regarding warnings, this pull request sets `-Wall` by default and includes an option for adding `-Wextra`. These options are fairly portable across compilers.


sayerhs:
@jrood-nrel I would recommend that you add `-pedantic` when `ENABLE_EXTRA_WARNINGS=ON`. We are already using `-std=c++11`, so it should flag any non-standard extensions being used that might not be uniformly supported across all compilers. Here is an example code that demonstrates this:

```
// File test.cpp
// Demonstrate use of -pedantic

#include <iostream>

double testfunc(const int n, double* vec1)
{
    double vec2[n];

    for(int i=0; i < n; i++)
        vec2[i] = 2.0 * vec1[i];

    double sum = 0.0;
    for(int i=0; i < n; i++)
        sum += vec2[i];

    return sum;
}

int main()
{
    constexpr int ndim=3;
    double vec1[ndim] = { 1.0, 2.0, 3.0};

    std::cout << "Sum is: " << testfunc(ndim, vec1) << std::endl;
    return 0;
}
```

**Compiler output (OS X LLVM/CLang)**

```
$ c++ -Wall -Wextra -std=c++11 test.cpp && ./a.out
Sum is: 12
$ c++ -Wall -Wextra -pedantic -std=c++11 test.cpp && ./a.out
test.cpp:6:16: warning: variable length arrays are a C99 feature
      [-Wvla-extension]
    double vec2[n];
               ^
1 warning generated.
Sum is: 12
```

sayerhs:
@jrood-nrel Have you tested this on Intel compilers? I think Intel uses `-wN` instead of `-Wextra` and `-Wpedantic`. You might want a compiler guard with the CMake options for that case?

jrood-nrel:
@sayerhs I tested the flags with `icc` and it doesn't error, and I researched warnings in general with Intel. As far as I know `-Wall -Wextra -pedantic` is fine. I'm not sure it uses `-pedantic`, but I'm pretty sure it uses `-Wall -Wextra` as expected, and at least adding these shouldn't cause problems when compiling with Intel.

PR #331, opened 2018-02-20
Bug fix for TurbKinEne Ksgs kernel, was missing coordinates_ definition.

jamelvin:
When I tried to use the source term defined in TurbKineticEnergyKsgsSrcElemKernel (this is not the DesignOrder one), I got an error when trying to add the master element call to SCV_VOLUME.  I believe this was due to the coordinates field not being defined, which this pull request resolves.


spdomin:
How is the milestoneRunConsolidated test running?

           turbulent_ke: [lumped_turbulent_ke_time_derivative, upw_advection_diffusion, ksgs]

spdomin:
Its probably due to the fact that your sim is missing lumped_mass. However, your path is appropriate as these kernels should be viewed as autonomous.

PR #336, opened 2018-03-05
Add Git commit information to log file output.

sayerhs:
This commit enables determination of Trilinos and Nalu Git commit SHA
IDs during CMake configuration phase and creates a new header
NaluVersionInfo.h with these parameters.

The version information is printed in the high-level banner in the log
file for tracking purposes. The version information is also printed when
the user invokes naluX executable with the '-v, --version' option.


sayerhs:
Sample output:

```
=================================================================
                            Nalu:
      A low Mach number, turbulent reacting flow code with
            coupling to PMR and object response (CHT)
=================================================================

   Nalu Version: v1.2.0-194-g54c9ef6
   Nalu GIT Commit SHA: 54c9ef62bd6283969338ad53b7a9d37bcab724f8
   Trilinos Version: 12.13-gac4d706

   TPLs: Boost, HDF5, netCDF, STK, Trilinos, YAML_cpp and zlib
   Optional TPLs enabled: OpenFAST, Hypre, TIOGA

              Copyright 2014 Sandia Corporation.
      This software is released under the license detailed
   in the file, LICENSE, which is located in the top-level Nalu
                     directory structure
-----------------------------------------------------------------
```

sayerhs:
@spdomin Yes, if Trilinos/Sierra team has a better way to handle the Git revision, let us use that. 

The one feature I want in the version embedding is that it detects a change in git commit/dirty status every time the user executes `make` and not just at CMake configure phase. This is why I am using the `GetGitRevisionDescription.cmake` instead of just grabbing the output of `git describe --tags` from the command line.

jrood-nrel:
One more comment I forgot is there seems to be a `GetGitRevisionDescription.cmake` and `GetGitRevisionDescription.cmake.in` file? I just didn't really understand why both exist and looked quite different.

sayerhs:
@jrood-nrel 

- I have added an update that will fix the issue with versions with `.tar.gz` use case. The mode of update of this string will be similar to what you have in `docs/conf.py` and will have to be manually updated whenever Nalu is tagged with a new version. The CMake functions inspect the `.git` directory instead of running `git describe`, so it should not have defined `NALU_HAVE_GIT_INFO` and, therefore, the `#else` section should have been used in `NaluVersionInfo.h`. 

- Take a look at `man git-describe` to see what the number + SHA version means, particularly the EXAMPLES section. `git describe` adds a `g` prefix to indicate it is a Git controlled repository as opposed to `hg (mercurial)`. I suspect this has something to do with having an uniform way to describe a version from different SCMs. I am not crazy about the `g` prefix, but my preference is to retain `git describe` convention than creating a Nalu-specific one. 

- As I mentioned in my comment to Stefan, I want the version refreshed/updated everytime the user executes `make` and not just when the user does `rm -rf CMakeFiles && cmake` step. So the `.in` version of the `GetGitRevisionDescription.cmake` is used to trigger this behavior by the `configure_file` action (see line 87 in `GetGitRevisionDescription.cmake`). 

-There is a mode of failure where the user has a `.git` repo, but doesn't have `git` installed/accessible on their path during CMake. I don't have a good idea how to resolve this situation.

jrood-nrel:
I have been working with Shreyas and I have tested all the cases I was concerned about and they work as I expect.

The version string makes sense to me now as the latest version tag found, along with the number of commits since that tag, and then a `g<SHA ID>` where `g` means git since projects can be using different source control applications.

I also think it will be difficult to avoid the complexity of logic to find the necessary information this is providing as I have seen this done in a few other large applications I am aware of and they seem close to this. This will also be great to have embedded in the executable sooner than later for the milestones. I think I'm going to merge this and if a better solution surfaces, we can move to it.

spdomin:
Let me reach out to the Sierra community and reply back. This may take a day or two. As such, I hope that we have time to query the community.

PR #337, opened 2018-03-05
Add skew symmetric operator for CVFEM advection (momentum and scalar)

spdomin:
* at all element-baed interior and open, consolidated and non, add
a notion of skew symmetry.

* add realm and solution options interface to skew symmetric specification.

* add test coverage

Usage:

skew_symmetric_advection:
  - velocity: yes
  - mixture_fraction: no

Notes:

a) this implementation includes saving off two types of interpolation
types for Kernels... I think that this is fine, however, something to
note in the design.


spdomin:
I will pursue the patch of elements unit test concept soon.

PR #338, opened 2018-03-05
Account for overset BC parts with check_for_missing_bcs

sayerhs:
Fixes error when the user enables "check_for_missing_bcs" in the input
file and overset BC is active. The fix adds overset side_set targets to
the realm.bcPartVec_ so that enforce_bc_on_exposed_faces will skip these
parts during the BC check step.

Updated two regression tests to provide coverage for this ability.

### DIFF
--- a/include/Realm.h
+++ b/include/Realm.h
@@ -517,7 +517,14 @@ class Realm {
   // mesh parts for all interior domains
   stk::mesh::PartVector interiorPartVec_;
 
-  // mesh parts for all boundary conditions
+  /** Vector holding side sets that have been registered with the boundary
+   * conditions in the input file.
+   *
+   * The member is intended to for use in Realm::enforce_bc_on_exposed_faces to
+   * check for "exposed surfaces" that might have not been assigned BCs in the
+   * input file.
+   *
+   */
   stk::mesh::PartVector bcPartVec_;
 
   // empty part vector should it be required
--- a/include/overset/OversetManagerSTK.h
+++ b/include/overset/OversetManagerSTK.h
@@ -81,6 +81,8 @@ public:
 
   virtual ~OversetManagerSTK();
 
+  virtual void setup();
+
   // main method called for initialization
   virtual void initialize();
   
--- a/include/overset/TiogaBlock.h
+++ b/include/overset/TiogaBlock.h
@@ -48,7 +48,7 @@ public:
 
   /** Setup block structure information (steps before mesh creation)
    */
-  void setup();
+  void setup(stk::mesh::PartVector&);
 
   /** Initialize mesh data structure (steps after mesh creation)
    */
--- a/include/overset/TiogaSTKIface.h
+++ b/include/overset/TiogaSTKIface.h
@@ -46,7 +46,7 @@ public:
 
   /** Setup block structure information (steps before mesh creation)
    */
-  void setup();
+  void setup(stk::mesh::PartVector&);
 
   /** Initialize mesh data structure (steps after mesh creation)
    */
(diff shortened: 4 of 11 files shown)


sayerhs:
@spdomin 

- Currently we don't use `Realm::bcPartVec_` for anything other than this check.

- I don't have a good answer to your question about keeping `bcPartVec_` true to its name. Overset surfaces being appended to `bcPartVec_` are the exposed surfaces of interior meshes, and skin_mesh is finding them as exposed faces. So from that perspective, it is a boundary surface/side_set and makes sense to add it to `bcPartVec_`. If you have a different suggestion, for how to avoid throwing an error, please let me know.

sayerhs:
@spdomin Sorry, I take that back about `bcPartVec_` usage. Can you comment on `RadiativeTransportEquationSystem::compute_bc_intensity`, I had not noticed that use case before.

spdomin:
The PMR example is a legit use case for when someone might want a full collection of bc parts to serve up to a selector. However, as I think you know, this is a local bcPartVec_ off of the PMR equation system.

Feel free to push and we can unravel this later should we find an instance when we truly need a proper bc vector. In the meantime, interpreting this data member as "all things exposed" for use in only checking for bc coverage is appropriate.

sayerhs:
@spdomin OK. I'll add a comment in `Realm.h` documenting the intent of `bcPartVec_` so that others are aware of how it is being used currently. Once I have that commit, I'll merge this. Thanks!

spdomin:
@sayerhs, this is no big deal, however, I think that the new comment is actually wrong. Your code push changed the definition of bcPartVec_ from a vector containing all boundary condition parts a user activated to a vector of all possibly known parts that are exposed that the user may or may not have been registered - the later being the case for overset.

I suppose that my time with @alanw0 pair programing provided me with a new comment philosophy:)

Feel free to leave or change. 

-  // mesh parts for all boundary conditions
 +  /** Vector holding side sets that have been registered with the boundary
 +   * conditions in the input file.

sayerhs:
I am not sure I understand. The user does explicitly call out which of the _exposed faces_ they want to treat as overset BC surfaces. For example, 

```
    - overset_boundary_condition: bc_left
      overset_user_data:
        percent_overlap: 10.0
        background_block: block_1
        overset_block: block_2
        overset_surface: surface_6                           #### Mark all nodes on this sideset as overset BC
        background_cut_block: block_3
        background_cut_surface: surface_101
```

```
    - overset_boundary_condition: bc_overset
      overset_connectivity_type: tioga
      overset_user_data:
        mesh_group:
          - overset_name: interior
            mesh_parts:
              - Unspecified-2-HEX
              - Unspecified-3-HEX
              - Unspecified-4-HEX
              - Unspecified-5-HEX
              - Unspecified-6-HEX
              - Unspecified-7-HEX
            wall_parts: [ wall ]
            ovset_parts: [ overset ]                             ### Mark all nodes on this sideset as overset BC

          - overset_name: exterior
            mesh_parts: [ Unspecified-8-HEX ]
```

spdomin:
Right. I guess that all is well. Thanks for the reminder of overset. I suppose that the skinned part is completed after the exposed check?

sayerhs:
@spdomin Yes, in the STK-based overset the `background_cut_surface` is created after `Realm::enforce_bc_on_exposed_faces` is called, both actions happen in `Realm::initialize`. The _part_ indicated by `background_cut_surface` is not added to `Realm::bcPartVec_`. 

The TIOGA approach does not skin mesh or create the equivalent of `background_cut_surface`.

spdomin:
Good to know. We can probably add that part to bcPartVec_ as well, however, this is probably not critical given that the usage of bcPartVec_ is for error checking only at present. We can file this away and add it later if we ever need such a part.

PR #344, opened 2018-03-12
Add sorting operation of exposed faces by ordinal

spdomin:
* NGP face/element approach requires a homogeneous collection
of ordinals served up for SIMD purposes. This push provides a
re-order of exposed faces by ordinal. At present, this sort is active
when the consolidated approach is active.

* activate sort only when symmetry is active and consolidated approach is in use.
This is a temp approach until we have full at-scale testing. The sort was demonstatred
to work in the dgThreeBlade case.

* re-bless threeBlade; consolidated with symmetry; non-consolidated restart also diffed.

* re-bless elemHybridShift; consolidated with symmetry again ~1e-12

Notes:

1) More detail is required to understand when custom ghosting affects face (non-ghosted)
ordering.


include/EntityExposedFaceSorter.h
```diff
@@ -0,0 +1,50 @@
...
+class EntityExposedFaceSorter : public stk::mesh::EntitySorterBase {
+
+  public:
+    
+  virtual void sort(stk::mesh::BulkData &bulk, stk::mesh::EntityVector& entityVector) const
+  {
+    stk::mesh::EntityRank entityVecRank = bulk.entity_rank(entityVector[0]);
+
+    if ( entityVecRank == stk::topology::FACE_RANK || entityVecRank == stk::topology::EDGE_RANK ) {
+      std::sort(entityVector.begin(), entityVector.end(),
```
alanw0: Do you want to apply this sort to edges? Or did you just include EDGE_RANK for 2-D meshes where edges are sides? To avoid sorting edges in a 3-D mesh, it might be better to use 'if (entityRank == bulk.mesh_meta_data().side_rank())'

spdomin: Yes, I added EDGE_RANK for the 2D use case. However, I can use the side_rank() just as easily. In general, this will not be active when we have edges in 3D - at least at present. I will make the change now.

spdomin: @alanw0, okay, I made the change. Feel free to squash these commits.

sayerhs:
@spdomin This is causing diffs in `dgncThreeBladeHypre`, @jrood-nrel will update the golds for that case also. Just an FYI.

spdomin:
Right... Sorry. I should probably update to include Hypre. These diffs are expected, however.

PR #346, opened 2018-03-13
Add mean resolved ke to turbulence averaging

spdomin:
* HIT and T-G sims require mean ke over the domain. This simple
commit promotes what I have been maintaining under LowMachEQS for
purposes of SNL NGS project.

* helium plume test modified to exercise the new post processing. Here,
the volume matches.

Notes:

a) In my -pi:pi^3 domain with triple peridocity, nodally iterating
provides a "periodic" volume as apposed to only the volume encompassed
by the domain. As such, the new postprocessing is actually more accurate
- at least for low-order.


src/TurbulenceAveragingPostProcessing.C
```diff
@@ -705,6 +710,14 @@ TurbulenceAveragingPostProcessing::execute()
       compute_lambda_ci(avInfo->name_, s_all_nodes);
     }
     
+    if ( avInfo->computeMeanResolvedKe_ ) {
+      // need locally owned and active nodes
+      stk::mesh::Selector s_locally_owned_nodes
+        = metaData.locally_owned_part() & stk::mesh::selectUnion(avInfo->partVec_);
+      //& !(realm_.get_inactive_selector());
```
rcknaus: I think this line shouldn't be commented out

spdomin: Yes, this was a WIP debug.. .I will fix and push.

PR #349, opened 2018-03-14
Add mean resolved ke to turbulence averaging

spdomin:
* HIT and T-G sims require mean ke over the domain. This simple
commit promotes what I have been maintaining under LowMachEQS for
purposes of SNL NGS project.

* helium plume test modified to exercise the new post processing. Here,
the volume matches.

* fixed a trivial name while I am here

* add inactive selector as per RCK review

Notes:

a) In my -pi:pi^3 domain with triple peridocity, nodally iterating
provides a "periodic" volume as apposed to only the volume encompassed
by the domain. As such, the new postprocessing is actually more accurate
- at least for low-order.

### DIFF
--- a/include/AveragingInfo.h
+++ b/include/AveragingInfo.h
@@ -45,11 +45,12 @@ public:
   bool computeVorticity_;
   bool computeQcriterion_;
   bool computeLambdaCI_;
+  bool computeMeanResolvedKe_;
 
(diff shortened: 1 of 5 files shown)


spdomin:
Sorry, that small name change in the Constraint class messed up everything...

PR #350, opened 2018-03-16
Add mean resolved ke to turbulence averaging; final bug fix

spdomin:
* The selector needs to include !slave_nodes... Now, volume is matching
the previous approach.


spdomin:
My apologies for the churn. I thought that this would be a quick change from elemental to nodal post processing, however, all of the different types of "inactive/slave" parts made the nodal loop more complex. @sayerhs, I would appreciate any thoughts that you might have on selector design points to insure that mistakes like I was making are less likely.

PR #351, opened 2018-03-16
First exposed face bc algorithm (consolidated)

spdomin:
* add new element types to BcAlgTraits

* add face only instantiation

* add RadTransWallElemKernel

* add kernel builder for face algs

* remove Built Supplemental Algs in favor of Kernels

* build template for face_bc

Notes:

a) good news, no diffs noted (concentic and inputFire)
b) ready to deploy for all other face-only bc algorithms.
c) sadly, AssembleRadTransWall is still there for edge-based only
d) new design looks nice and clean.

### DIFF
--- a/include/BcAlgTraits.h
+++ b/include/BcAlgTraits.h
@@ -25,6 +25,43 @@ struct BcAlgTraitsHex8Quad4 {
   static constexpr stk::topology::topology_t elemTopo_ = stk::topology::HEX_8;
 };
 
+struct BcAlgTraitsQuad4 {
+  static constexpr int nDim_ = 3;
+  static constexpr int nodesPerFace_ = 4;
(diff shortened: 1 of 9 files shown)


spdomin:
We need to see if @rcknaus can make an incremental push (after this one) to deal with face/element pairs. At present, I have BcAlgTraits that are face-only and one hard coded Hex8Quad4 one (for the Symmetry/Open unit test). 

I would have changed it all, however, want to avoid clashing with Robert's work.

spdomin:
Right. For the simple face-only algorithm paradigm, the ordinal complexity is mitigated. The concept of sharing ElemSolver really worked out nicely. Thanks for the help. We can convert face-only algorithms at will now. I plan on hitting the wall function approach next.

Issue #335, opened 2018-03-05
meSCV::grad_op() for promoted higher-order elements has not been implemented


spdomin:
The promoted P>1 elements do not have meSCV::grad_op() implemented. This is causing my design-order Ksgs runs to throw.

Could one of you take this on? If you do not have the cycles, let me know and I can find the time.

marchdf:
@spdomin I am focused on another ECP milestone this month so I don't think I will be able to get to it in the next couple of weeks. If this can wait that long, I can take a stab.

spdomin:
Okay, we can see how @rcknaus availability is. In general, there is no huge rush. Moreover, this may fit well within FY18/Q4's ECP milestone.

PR #353, opened 2018-03-20
Add SCV grad_op for promoted elements

rcknaus:
* Addresses #335
* Gets rid of the unused ability to specify sub/super parametric elements for promoted elements.  
* Adds clipping for ip values in design order ksgs, since the interpolation isn't necessarily monotonic

### DIFF
--- a/src/kernel/TurbKineticEnergyKsgsDesignOrderSrcElemKernel.C
+++ b/src/kernel/TurbKineticEnergyKsgsDesignOrderSrcElemKernel.C
@@ -130,8 +130,12 @@ TurbKineticEnergyKsgsDesignOrderSrcElemKernel<AlgTraits>::execute(
         Pk += w_dudx[i][j]*(w_dudx[i][j] + w_dudx[j][i]);
       }
     }
-    Pk *= tviscIp;
-    
+    Pk *= stk::math::max(tviscIp,0);
+    tkeIp = stk::math::max(tkeIp,0);
+    dualNodalVolIp = stk::math::max(dualNodalVolIp,0);
+    rhoIp = stk::math::max(rhoIp,0);
+
+
     // tke factor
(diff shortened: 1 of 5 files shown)


src/kernel/TurbKineticEnergyKsgsDesignOrderSrcElemKernel.C
```diff
@@ -130,8 +130,12 @@ TurbKineticEnergyKsgsDesignOrderSrcElemKernel<AlgTraits>::execute(
         Pk += w_dudx[i][j]*(w_dudx[i][j] + w_dudx[j][i]);
       }
     }
-    Pk *= tviscIp;
-    
+    Pk *= stk::math::max(tviscIp,0);
+    tkeIp = stk::math::max(tkeIp,0);
+    dualNodalVolIp = stk::math::max(dualNodalVolIp,0);
+    rhoIp = stk::math::max(rhoIp,0);
```
spdomin: Can we clip this to something more reasonable? I think that, at present, we only have the ability to clip Z to a reference value. However, perhaps this is not required. 

I guess that I am surprised that rho and dualNodalVolIp need clipping - especially since they should not hover close to 0.0. Have you seen this in tests or is this just a worst case approach?

spdomin:
Until we resolve the clipping approach, I would appreciate if we did not push.

marchdf:
@rcknaus wondering if you need any help with this? To get my pull request #366 in, it looks like I need to implement `SCV_SHIFTED_GRAD_OP` (so I can get `dndx_scv_shifted`). I can work on doing that and maybe we can use this to make it work for high-order as well (assuming things need to be adapted). Let me know what you think.

rcknaus:
Clipping density and the dual volume don't seem necessary.  I just added them as a worst case approach. For this type of clipping, I think we do need to decide on a general strategy for how we want to handle it.

PR #355, opened 2018-03-22
(WIP) Prototype a utility kernel design

rcknaus:
* Adds the concept of a "utility kernel" that takes some input fields and writes to some other output fields
* Implemented for just pure nodal loops.  
* Changed the turbviscsmagorinsky algorithm to run the kernel.  It's not how it would be implemented in the end---it's just an easy way to check if it the new version causes regressions / passes the unit test.

### DIFF
--- a/include/NodalScratchData.h
+++ b/include/NodalScratchData.h
@@ -1,0 +1,101 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 Sandia Corporation.                                    */
+/*  This software is released under the license detailed                  */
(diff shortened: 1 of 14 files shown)


sayerhs:
@rcknaus Overall, I think the code is good. I have some questions regarding this  agree with @spdomin that a walk through will be good if you are willing to set a meeting. If you disagree, then I will post my questions here. 

Not doing a review since this is tagged (WIP).

spdomin:
Push to Nalu-Wind.

PR #356, opened 2018-03-22
New face/element wall function kernel

spdomin:
* converted Wall function for element-based scheme. The prior AssembleElemWallFuncSolverAlg implementation is still active when non-consolidated approach is activated. This will allow for
continued performance testing of consolidated vs non-consolidated. However, we are getting close to deprecating the full element-based approach to use Kokkos.

* add new enums (yplus_crit, elog) to avoid hard coding in the kernel. These
are still constant in time and space.

Notes:

a) reporting needs to be cleaned up; will add a story for that.
b) ctest showed no diffs while there did seem to be some at the 1e-17
level... re-blessed even though it passed.


spdomin:
Yes, really liking the look and feel of BC kernels.

PR #360, opened 2018-03-29
Subtract out normal pressure gradient when mdot correction BC is active

rcknaus:
* adds a solution option:  explicitly_zero_open_pressure_gradient: yes 
* explicitly subtracts out the normal component of the pressure gradient at an open boundary, meant to be used with the mdot_correction routine
* Adds a capping inversion user function to make testing atmospheric boundary layer flows a bit more straightforward
* Changed ablUnstableEdge_rst to use new functionality and reblessed the gold file


spdomin:
Open mean normal is not correct for cases in which periodic

spdomin:
Closing. Feel free to open under nalu-wind.

PR #361, opened 2018-03-30
Hypre: Use new functions in Hypre API to zero matrix.

sayerhs:
Prior to Hypre v2.14.0, there was no quick way to zero out the entries
in an IJMatrix while preserving the connections. Nalu interface to Hypre
had resorted to destroying and recreating the IJMatrix during the call
to HypreLinearSystem::zeroSystem. This meant that the IJMatrix was
recreated during every outer iteration even if there was no mesh motion.
This overhead is removed by using HYPRE_IJMatrixSetConstantValues to
zero out entries but preserve the matrix structure from previous calls.

### DIFF
--- a/src/HypreLinearSystem.C
+++ b/src/HypreLinearSystem.C
@@ -333,16 +333,11 @@ HypreLinearSystem::zeroSystem()
   MPI_Comm comm = realm_.bulk_data().parallel();
   HypreDirectSolver* solver = reinterpret_cast<HypreDirectSolver*>(linearSolver_);
 
-  if (systemInitialized_)
-    HYPRE_IJMatrixDestroy(mat_);
-  HYPRE_IJMatrixCreate(comm, iLower_, iUpper_, jLower_, jUpper_, &mat_);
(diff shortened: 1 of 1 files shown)


sayerhs:
@shashankNREL I am requesting you to look at the code to make sure the Hypre implementation is correct. Let @jrood-nrel be the person approving/merging this as this requires latest Hypre updates to Spack installs. 

@jrood-nrel This requires the latest `master` or `v2.14.0` from Hypre that was released a few days ago. The update seemed straightforward at my end, so let me know if you run into issue.

jrood-nrel:
I've already told @sayerhs this, but I am waiting to see what happens with this pull request https://github.com/spack/spack/pull/7606 before implementing Hypre 2.14.0 in the tests and merging this. If that pull request takes too long, I can also just implement it myself on the test machines though so we can try to merge this in a timely manner.

sayerhs:
@jrood-nrel This is ready to be merged now that Hypre 2.14.0 is in the Spack builds. @sthomas61 has tested this out on our problem sets.

PR #362, opened 2018-04-05
Add logic to repeat search at non-conformal interfaces.

overfelt:
If the search fails to find a match for some elements then
increase the search distance and repeat the search with just
the elements that were not matched.  This will keep the dynamic
search tolerance algorithm from failing if the search tolerance
suddenly needs to be increased by a large factor from one time
step to the next. If after doubling the search tolerance ten times
and there are still points with no match the function then throws
an error.


spdomin:
@overfelt, I will have a detailed look at this later today or Monday. Many thanks for the PR..

PR #364, opened 2018-04-13
Optimization of Tpetra init (construction of graphs etc)

alanw0:
Sorry about putting in such a large code-change. If it helps I can split it into smaller pieces for easier review...

Add communication code to send/recv off-processor column indices
so that we can construct Tpetra graphs with exact row-lengths and
use the 'StaticProfile' constructor argument which avoids significant
re-allocation during fillComplete.
Also eliminated temporary graph which was being used to import
remove column indices for the owned graph.

Had to temporarily take copies of stk files CommBufferV.hpp and
CommNeighbors.[hpp,cpp], until next time stk is updated. I made some
important performance improvements in these files. For now, they
reside in include/NaluCommBufferV.hpp, include/NaluCommNeighbors.hpp and
src/NaluCommNeighbors.C.

Added ghosting modification to PeriodicManager to ensure that ghosts
of range nodes are on the procs that share the domain nodes.

Added tpetra unit-test, which required enhancing UnitTestRealm.C to
add some YAML-node input syntax adding an equation system etc.

Had to compute local graph arrays (row-pointers and col-indices) to
give to Tpetra using setAllIndices(..) to avoid having Tpetra sort
all the rows during fillComplete.

Tried using expertStaticFillComplete. Went back to
regular fillComplete, as the 'expert' one doesn't perform any
better than regular. Consulted with Mark Hoemmen, he says the
expert one accepts a pre-constructed Importer which could save time
if we can pre-construct an Importer with our provided remote-owner
information so that Importer doesn't have to communicate. He has
just added a function to construct an Importer that way, so we can
switch to that after Trilinos updates work their way through.

Reblessed 19 tests due to changed ordering of matrix graph. All diffs
were consistent with round-off as demonstrated by Robert's experiment
of re-running tests with changed stk-mesh ordering.

Performance result: some nightly tests such as uqSlidingMeshDG and
waleElemXflowMixFrac3.5m show improvements in tpetra initialization time
around 2x. The V27 problem running on 8192cores of cori (setup courtesy of
Shreyas and Jonathan Hu) showed init time decrease from ~116s to ~75s, with
overall runtime decrease from ~375s to ~345s.


sayerhs:
@alanw0 It appears that unit tests depending on TIOGA and OpenFAST have not been updated, is this just fortunate or were those libraries not enabled in your Nalu build? Since all other overset tests and ABL tests changed, I would suspect that these tests will require updates of gold files also.

alanw0:
@sayerhs my build would have only included what's enabled by default. Do I need to install TPLs to enable TIOGA and OpenFAST?

sayerhs:
@alanw0 Yes, that is correct. You will have to install TIOGA and OpenFAST and enable them during CMake configure phase.

alanw0:
@sayerhs do you have a recipe for obtaining/installing/enabling those TPLs? Sorry, I don't see anything obvious although I haven't done much google-ing yet.

sayerhs:
@alanw0 I'll email you the instructions.

alanw0:
@sayerhs ok. I found tioga on GitHub and cloned it and built it, but when I try to enable it in the nalu build I get errors about "By not providing "FindTIOGA.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "TIOGA", but
  CMake did not find one.".
So I'll wait for your email before proceeding further with this.

include/LinearSolver.h
```diff
@@ -54,6 +55,79 @@ namespace nalu{
...
+  LocalGraphArrays(const Kokkos::View<size_t*,sierra::nalu::DeviceSpace>& rowLengths)
+  : rowPointers(Teuchos::arcp<size_t>(rowLengths.size()+1)),
+    colIndices()
+  {
+    size_t nnz = compute_row_pointers(rowPointers, rowLengths);
+    colIndices = Teuchos::arcp<LocalOrdinal>(nnz);
+    for(int i=0, iend=colIndices.size(); i<iend; ++i) { colIndices[i] = INVALID; }
+  }
+
+  size_t get_row_length(size_t localRow) const { return rowPointers[localRow+1]-rowPointers[localRow]; }
```
mhoemmen: You're assuming UVM here.

include/LinearSolver.h
```diff
@@ -54,6 +55,79 @@ namespace nalu{
...
+
+  void insertIndices(size_t localRow, size_t numInds, const LocalOrdinal* inds, int numDof)
+  {
+    LocalOrdinal* row = &colIndices[rowPointers[localRow]];
```
mhoemmen: Use Kokkos::View for colIndices, and take a subview.

include/LinearSolver.h
```diff
@@ -54,6 +55,79 @@ namespace nalu{
...
+    size_t rowLen = rowPointers[localRow+1] - rowPointers[localRow];
```
mhoemmen: You have a function that does this; why not use it?

include/LinearSolver.h
```diff
@@ -54,6 +55,79 @@ namespace nalu{
...
+    LocalOrdinal* rowEnd = find_first_invalid(row, rowLen);
+    for(size_t i=0; i<numInds; ++i) {
+      LocalOrdinal* insertPoint = std::lower_bound(row, rowEnd, inds[i]);
```
mhoemmen: You could also use the following Trilinos function:

https://github.com/trilinos/Trilinos/blob/a27130bc4f5d7dba02be3cb976d1edc8653f941b/packages/kokkos-kernels/src/sparse/KokkosSparse_findRelOffset.hpp#L114

include/LinearSolver.h
```diff
@@ -54,6 +55,79 @@ namespace nalu{
...
+      if (insertPoint <= rowEnd && *insertPoint != inds[i]) {
+        insert(inds[i], numDof, insertPoint, rowEnd+numDof);
+        rowEnd += numDof;
+      }
+    }
+  }
+
+  static size_t compute_row_pointers(Teuchos::ArrayRCP<size_t>& rowPtrs,
```
mhoemmen: 1. Tpetra has functions for computing row offsets.  They are thread-parallel, work on the GPU, etc.  If you don't want to use them, at least use Kokkos::parallel_scan.

2. Don't use ArrayRCP (see above).

include/NaluCommBufferV.hpp
```diff
@@ -0,0 +1,122 @@
...
+        if (num_bytes > (data_buffer.capacity() - data_buffer.size())) {
+            data_buffer.reserve(std::max(num_bytes, data_buffer.capacity()*2));
+        }
+        const unsigned char* item_chars = reinterpret_cast<const unsigned char*>(items);
+        data_buffer.insert(data_buffer.end(), item_chars, item_chars+num_bytes);
+        unpack_iterator = data_buffer.begin();
+    }
+
+    std::vector<unsigned char> data_buffer;
+    std::vector<unsigned char>::iterator unpack_iterator;
```
mhoemmen: Hm, why not write this in a more functional way, by returning the iterator from pack/unpack methods instead of keeping it in the class?

include/NaluCommNeighbors.hpp
```diff
@@ -0,0 +1,133 @@
...
+#include <cstddef>                      // for size_t, ptrdiff_t
+#include <vector>
+#include <stk_util/parallel/Parallel.hpp>  // for ParallelMachine
+#include <NaluCommBufferV.hpp>
+
+//------------------------------------------------------------------------
+
+namespace nalu_stk {
+
+class CommNeighbors {
```
mhoemmen: This looks suspiciously like yet another boundary exchange implementation.

MPI 3 has a boundary exchange implementation.  Dan Ibanez has another one (that wraps / looks like MPI 3's interface).

mhoemmen: What I'm saying is, why reimplement?

src/NaluCommNeighbors.C
```diff
@@ -0,0 +1,156 @@
...
+  const int mpitag = 10101, mpitag2 = 10102;
+  int maxRecvProcs = m_recv_procs.size();
+  int maxSendProcs = m_send_procs.size();
+  int max = std::max(maxSendProcs, maxRecvProcs);
+  std::vector<int> recv_sizes(maxRecvProcs, 0);
+  std::vector<MPI_Request> requests(max, MPI_REQUEST_NULL);
+  std::vector<MPI_Request> requests2(max, MPI_REQUEST_NULL);
+  std::vector<MPI_Request> requests3(max, MPI_REQUEST_NULL);
+  std::vector<MPI_Status> statuses(max);
+  for(int i=0; i<maxRecvProcs; ++i) {
```
mhoemmen: My suspicion was correct: this is another boundary exchange reimplementation.

Use Tpetra::Distributor, or Dan Ibanez's interface, or MPI 3's, or maybe STK's (because surely it has one).  Don't reimplement this.  Boundary exchange implementations may need to change a LOT in the next few years for good performance.  The fewer reimplementations we have to rewrite, the better.

alanw0: @mhoemmen you must be referring to MPI_Neighbor_alltoallv. I'll investigate that, thanks.

mhoemmen: Tpetra's doesn't use a global sync.  Neither does the MPI 3 version, and I doubt that Dan's does either.

See Chapter 7 of the MPI 3.0 Standard for neighborhood collectives and process topologies.  You would use `MPI_Dist_graph_create_adjacent` to create the communication pattern (see Example 7.3, pp. 300-1, esp. the last table in the example).  The pattern gets attached to a new `MPI_Comm` which is otherwise a clone of the input `MPI_Comm`.  You then pass the resulting `MPI_Comm` to either of the following:

  - `MPI_Neighbor_alltoall` (see Section 7.6.2, p. 319), if you need to send a fixed number of things to/from each process, or
  - `MPI_Neighbor_alltoallv` (p. 321), if you need to send a variable number of things to/from each process (which should be the case here)

The corresponding `Tpetra::Distributor` pattern would be: 
  1. Create Distributor using `createFromSendsAndRecvs`
  2. Use either three-argument doPostsAndWaits (for the "alltoall" case above), or four-argument doPostsAndWaits (for the "alltoallv" case above)

alanw0: @mhoemmen thanks. I'll confess I didn't know the details of Tpetra::Distributor, and have never heard of Dan's. The MPI one wasn't available to us in sierra until we upgraded to open-mpi 1.10. (Which is still not new, but at least it has the MPI_Neighbor_* functions.)

mhoemmen: @alanw0 Dan's isn't in Trilinos (yet! ;-) ).

`Tpetra::Distributor` (and `Epetra_*Distributor`) also implement boundary exchange.  They are the communication layer that doExport and doImport use underneath.

alanw0: @mhoemmen  I've pushed another commit, addressing your comments (removing ArrayRCP in favor of View, and using MPI_Neighbor boundary exchange).

spdomin:
I like Mark's comments:) Will have a look this week.

include/LinearSolver.h
mhoemmen: Be careful w/ "HostSpace" -- if that's a memory space, its default execution space might not be the same as the default execution space you intend to use.

You could leave these as device Views -- Tpetra has a "getEntryOnHost" function that you could copy for this purpose.

include/LinearSolver.h
mhoemmen: The function I mentioned above works on device too.

include/LinearSolver.h
```diff
@@ -90,20 +91,22 @@ class LocalGraphArrays {
     }
   }
 
-  static size_t compute_row_pointers(Teuchos::ArrayRCP<size_t>& rowPtrs,
-                                   const Kokkos::View<size_t*,sierra::nalu::DeviceSpace>& rowLengths)
+  static size_t compute_row_pointers(Kokkos::View<size_t*>& rowPtrs,
```
mhoemmen: The device type is missing; did you mean to do that?

include/LinearSolver.h
```diff
@@ -90,20 +91,22 @@ class LocalGraphArrays {
...
-      rowPtrs[i] = nnz;
-      nnz += rowLengths(i);
+    size_t* rowPtrData = rowPtrs.data();
+    const size_t* rowLens = rowLengths.data();
+    for(unsigned i=0, iend=rowLengths.size(); i<iend; ++i) {
+      rowPtrData[i] = nnz;
+      nnz += rowLens[i];
     }
-    rowPtrs[rowLengths.size()] = nnz;
+    rowPtrData[rowLengths.size()] = nnz;
```
mhoemmen: See above comments about device Views.

include/LinearSolver.h
```diff
@@ -90,20 +91,22 @@ class LocalGraphArrays {
...
     return nnz;
   }
 
-  Teuchos::ArrayRCP<size_t> rowPointers;
-  Teuchos::ArrayRCP<LocalOrdinal> colIndices;
+  Kokkos::View<size_t*> rowPointers;
```
mhoemmen: Is it OK to use the default device type here?

include/NaluCommBufferV.hpp
```diff
@@ -67,6 +67,7 @@ class CommBufferV {
     }
 
     unsigned char* raw_buffer() { return data_buffer.data(); }
```
mhoemmen: Any particular reason why this is `unsigned char`?  `char` seems to work fine here.

Also, why not just call this method `data()`?

include/NaluCommBufferV.hpp
```diff
@@ -67,6 +67,7 @@ class CommBufferV {
...
+    const unsigned char* raw_buffer() const { return data_buffer.data(); }
```
mhoemmen: See above comment.

src/NaluCommNeighbors.C
```diff
@@ -84,7 +85,17 @@ CommNeighbors::CommNeighbors( stk::ParallelMachine comm, const std::vector<int>&
...
   m_send.resize(m_size);
   m_recv.resize(m_size);
-  std::sort(m_neighbor_procs.begin(), m_neighbor_procs.end());
+  stk::util::sort_and_unique(m_send_procs);
+  stk::util::sort_and_unique(m_recv_procs);
+  MPI_Info info;
+  MPI_Info_create(&info);
+  int reorder = 0;
+  const int* weights = (int*)MPI_UNWEIGHTED;
+  MPI_Dist_graph_create_adjacent(comm,
```
mhoemmen: It might still make sense to wrap this function with an interface, so you're not committed to it.  I haven't benchmarked MPI neighborhood collectives.  They _should_ be optimized.

src/NaluCommNeighbors.C
```diff
@@ -102,52 +113,91 @@ CommNeighbors::CommNeighbors( stk::ParallelMachine comm, const std::vector<int>&
   m_recv.resize(m_size);
   stk::util::sort_and_unique(m_send_procs);
   stk::util::sort_and_unique(m_recv_procs);
+  MPI_Info info;
+  MPI_Info_create(&info);
+  int reorder = 0;
+  const int* weights = (int*)MPI_UNWEIGHTED;
+  MPI_Dist_graph_create_adjacent(comm,
```
mhoemmen: See above comment.

include/NaluCommNeighbors.hpp
```diff
@@ -122,7 +122,6 @@ class CommNeighbors {
   std::vector<unsigned char> m_recv_data;
   std::vector<int> m_send_procs;
   std::vector<int> m_recv_procs;
```
mhoemmen: If you use MPI neighborhood collectives, you don't actually need to keep the send and receive process ranks.  MPI keeps them for you.  MPI has functions for querying a topology, so you can get this information if you need it.

src/TpetraLinearSystem.C
```diff
@@ -773,7 +773,7 @@ void communicate_remote_columns(const stk::mesh::BulkData& bulk,
                                 nalu_stk::CommNeighbors& commNeighbors,
                                 unsigned numDof,
                                 const Teuchos::RCP<LinSys::Map>& ownedRowsMap,
-                                Kokkos::View<size_t*,DeviceSpace>& deviceLocallyOwnedRowLengths,
+                                Kokkos::View<size_t*,HostSpace>& deviceLocallyOwnedRowLengths,
```
mhoemmen: See above comment on HostSpace.

src/TpetraLinearSystem.C
```diff
@@ -909,8 +909,8 @@ TpetraLinearSystem::compute_graph_row_lengths(const ConnectionVec& connectionVec
                                               LinSys::RowLengths& locallyOwnedRowLengths,
                                               nalu_stk::CommNeighbors& commNeighbors)
 {
-  Kokkos::View<size_t*,DeviceSpace> deviceGloballyOwnedRowLengths = globallyOwnedRowLengths.view<DeviceSpace>();
-  Kokkos::View<size_t*,DeviceSpace> deviceLocallyOwnedRowLengths = locallyOwnedRowLengths.view<DeviceSpace>();
+  Kokkos::View<size_t*,HostSpace> deviceGloballyOwnedRowLengths = globallyOwnedRowLengths.view<HostSpace>();
```
mhoemmen: See above comment on HostSpace.

include/LinearSolver.h
alanw0: I should probably specify device type for all views... In this case, rowPointers and colIndices are for passing into CrsGraph::setAllIndices, which takes these two types:
   const typename local_graph_type::row_map_type & rowPointers
   const typename local_graph_type::entries_type::non_const_type & columnIndices
I've had trouble finding exactly what those types need to be. Do they expect a particular device type? Can setAllIndices (and other CrsGraph methods) be called on either host or device (in gpu context)?
For now, our linear-system assembly is tied to the host, but in the future it may not be.

src/NaluCommNeighbors.C
alanw0: Yes some extra wrapping might be a good idea here. I was worried about performance, but I ran these changes on the 8192-rank case on the cori machine a few times over the weekend and performance seems to be ok. But of course that doesn't speak for all MPI implementations on all platforms...

include/NaluCommNeighbors.hpp
alanw0: Yes, that's true... If I do want to allow people to get them, I think hanging onto the vectors is a better solution. I know MPI provides functions for getting the neighbor-counts and the neighbors, but then I have to make the counts call, allocate the vector, then make the second call... I'm just not sure yet about the use-case for a caller needing them. That probably means I should remove the access entirely until the need becomes clear.

include/LinearSolver.h
mhoemmen: @alanw0 wrote: 
> I've had trouble finding exactly what those types need to be.

Those are Kokkos::View specializations.  `local_graph_type` is a Kokkos::StaticCrsGraph specialization.  (Tpetra::CrsGraph declares the `local_graph_type` typedef.)

> Do they expect a particular device type?

Their device type comes from the Tpetra class' `device_type` typedef.

> Can setAllIndices (and other CrsGraph methods) be called on either host or device (in gpu context)?

  1. Tpetra class functions are currently all host only.
  2. `setAllIndices` is a coarse-grained operation and therefore to be called on host.

mhoemmen: @alanw0 wrote:
> For now, our linear-system assembly is tied to the host, but in the future it may not be.

If you intend to fill on host, you could use host Views here or even host mirror views.  Host mirror views would be `CudaUVMSpace` currently, so please be sure to fence after filling on host.

src/NaluCommNeighbors.C
mhoemmen: @alanw0 If you do wrap this, please consider an interface that takes offsets rather than (or in addition to) counts.  That would give implementations the option to thread-parallelize underneath.

include/NaluCommNeighbors.hpp
mhoemmen: I agree that removing access for now is a good idea :-) .

include/LinearSolver.h
alanw0: @mhoemmen thanks for the info. given that Tpetra class functions are host only, I will use HostSpace for all of our views. Is there a plan for Tpetra functions to be device functions in the future? The stk mesh project has been experimenting with a 'ngp::Mesh' class which makes a device snapshot of the mesh, and has device functions for doing common mesh traversal/access operations. That's probably what nalu will use if/when we attempt to fill on device.

mhoemmen: @alanw0 wrote:
> given that Tpetra class functions are host only, I will use HostSpace for all of our views.

`setAllIndices` is a host function, yes, but it takes device Views.

mhoemmen: @alanw0 wrote:
> Is there a plan for Tpetra functions to be device functions in the future?

KokkosSparse::CrsMatrix has much of the functionality you need for sparse matrix assembly (with fixed graph structure -- sumInto/replace, not insert).  See e.g., `sumIntoValues` and `replaceValues`.

We have some plans to write easier-to-use graph construction interfaces, but for now, please see the Kokkos FENL example for thread-parallel graph construction.

spdomin:
@alanw0, let's find time to walk through this code change so that we can have more eyes on the details.

jhux2:
> Consulted with Mark Hoemmen, he says the
> expert one accepts a pre-constructed Importer which could save time
> if we can pre-construct an Importer with our provided remote-owner
> information so that Importer doesn't have to communicate. He has
> just added a function to construct an Importer that way, so we can
> switch to that after Trilinos updates work their way through.

@alanw0 @mhoemmen's new Import constructor is in master, see commit trilinos/Trilinos@43094b41 and the related issue trilinos/Trilinos#2198.

spdomin:
We plan on processing this PR today... Thanks to @alanw0, @mhoemmen, and @jhux2 for valuable assistance.

alanw0:
Good discussion this morning. Added a commit to change "globally owned" to "shared not owned".

jhux2:
@sthomas61 Fyi.

PR #365, opened 2018-04-16
Add jacobian check for mesh quality

spdomin:
* Within Realm scope, a new line command has been added:

    check_jacobians: yes/no

If this is active, compute_geometry() adds a loop over physics parts
to provide the list of bad elements (through global id).

* modified a test for an example

### DIFF
--- a/src/ComputeGeometryAlgorithmDriver.C
+++ b/src/ComputeGeometryAlgorithmDriver.C
@@ -11,6 +11,7 @@
 #include <Algorithm.h>
 #include <AlgorithmDriver.h>
 #include <FieldTypeDef.h>
+#include <master_element/MasterElement.h>
 #include <Realm.h>
 
 // stk_mesh/base/fem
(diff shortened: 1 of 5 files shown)


src/ComputeGeometryAlgorithmDriver.C
```diff
@@ -135,7 +136,106 @@ ComputeGeometryAlgorithmDriver::post_work()
...
+          for ( int j=0; j < nDim; ++j ) {
+            p_coordinates[niNdim+j] = coords[j];
+          }
+        }
+        
+        double scv_error = 0.0;
+        meSCV->determinant(1, &p_coordinates[0], &p_scv_volume[0], &scv_error);
+        
+        bool localBadElem = false;
+        for ( int ip = 0; ip < numScvIp; ++ip ) {
```
sayerhs: @spdomin Is there a reason to loop over and check `scv_volume` again instead of checking the value of `scv_error`?

src/ComputeGeometryAlgorithmDriver.C
```diff
@@ -135,7 +136,106 @@ ComputeGeometryAlgorithmDriver::post_work()
     realm_.periodic_field_update(dualNodalVolume, fieldSize);
   }
 
+  if ( realm_.checkJacobians_ ) {
+    check_jacobians();
```
sayerhs: Right now this step will be executed at every timestep when there is mesh motion, but no mesh deformation or percept refinement. Should we have a flag disabling this check after the initialization phase?

src/ComputeGeometryAlgorithmDriver.C
spdomin: I do not believe that scv_error is to be trusted. I can look through the ME implementation to see.

src/ComputeGeometryAlgorithmDriver.C
spdomin: Yes, my intent was to always check when mesh motion is active. I can probably add an extra logic statement on mesh deformation. Sadly, this would include some sort of flag management.

PR #366, opened 2018-04-17
Added SST source kernels and unit tests.

marchdf:
Most of the code here has already been reviewed (a while back...) but I finally got around to running a comparison using the current CVFEM implementation and a consolidated implementation that uses the new source kernels. The comparison was done using the NASA wall hump and a sample velocity profile can be seen in this graph. Both implementations predict the same velocity fields (same goes for surface coefficients).
u_2

The input files used for the comparison are the following:
- [previous implementation (using SuppAlg)](https://github.com/marchdf/wallHump/blob/master/twod/817x217/wallHump_p1.i)
- [using the new kernels (consolidated approach)](https://github.com/marchdf/wallHump/blob/master/twod/817x217/wallHump_p1_consolidated.i)

Appropriate unit tests were added for the new source kernels.

### DIFF
--- a/include/kernel/SpecificDissipationRateSSTSrcElemKernel.h
+++ b/include/kernel/SpecificDissipationRateSSTSrcElemKernel.h
@@ -54,6 +54,7 @@ private:
   VectorFieldType* coordinates_{nullptr};
 
   const bool lumpedMass_;
+  const bool shiftedGradOp_;
   const double betaStar_;
   const double sigmaWTwo_;
   const double betaOne_;
--- a/include/kernel/SpecificDissipationRateSSTSrcElemKernel.h
+++ b/include/kernel/SpecificDissipationRateSSTSrcElemKernel.h
@@ -1,0 +1,75 @@
+/*------------------------------------------------------------------------*/
+/*  Copyright 2014 National Renewable Energy Laboratory.                  */
+/*  This software is released under the license detailed                  */
+/*  in the file, LICENSE, which is located in the top-level Nalu          */
+/*  directory structure                                                   */
+/*------------------------------------------------------------------------*/
+
+#ifndef SPECIFICDISSIPATIONRATESSTSRCELEMKERNEL_H
+#define SPECIFICDISSIPATIONRATESSTSRCELEMKERNEL_H
+
+#include "Kernel.h"
+#include "FieldTypeDef.h"
+
+#include <stk_mesh/base/Entity.hpp>
+
+#include <Kokkos_Core.hpp>
+
+namespace sierra {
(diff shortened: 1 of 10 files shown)


marchdf:
@spdomin thank you for the comments. That is a good idea. It is my understanding that this would require adding a `SCV_SHIFTED_GRAD_OP` capability (similar to the existing `SCS` grad op) so that I could add something like this to those four kernels:
```
if (shiftedGradOp_)
    dataPreReqs.add_master_element_call(SCV_SHIFTED_GRAD_OP, CURRENT_COORDINATES);
else
    dataPreReqs.add_master_element_call(SCV_GRAD_OP, CURRENT_COORDINATES);
```
and
```
SharedMemView<DoubleType***>& v_dndx =
    shiftedGradOp_ ? scratchViews.get_me_views(CURRENT_COORDINATES).dndx_scv_shifted
                                   : scratchViews.get_me_views(CURRENT_COORDINATES).dndx_scv;
```
Is this a correct interpretation of your comment?

spdomin:
@marchdf, yes, that is correct. Logic both in the constructor and kernel.

We are still waiting on the PR from @rcknaus on the SCV::grad_op for generalized higher-order. If you two can work together to push this forward, that would be great. I think Robert's PR is active and really close to being accepted (it may have only had some clipping in need of clarification).

spdomin:
Looks like all of the scv::grad_op flavors are complete. Feel free to add the shifting interface and then process the PR.

marchdf:
@spdomin I pushed the shifting interface. If it looks good to you, I will process the PR. Thanks for your help.

spdomin:
@marchdf, I seem to be noting some unit test fails with a fresh pull/build?

[  FAILED  ] 3 tests, listed below:
[  FAILED  ] SSTKernelHex8Mesh.turbkineticenergysstsrcelem
[  FAILED  ] SSTKernelHex8Mesh.turbkineticenergysstdessrcelem
[  FAILED  ] SSTKernelHex8Mesh.specificdissipationratesstsrcelem

marchdf:
@spdomin seems like there is an problem with those tests when running with more than 1 proc (it passes with one proc). I will look into this asap. Thank you for pointing that out.

marchdf:
@spdomin The reason the tests fail with nprocs > 1 is that they are using 3D tests functions (they include a z-component). With nprocs > 1, kernels operate on elements that have different z-coordinates and therefore the golds fail. In accordance to what is currently being done for some tests, I intend on adding the following to all my tests:
```
if (stk::parallel_machine_size(MPI_COMM_WORLD) > 1)
    return;
```

PR #367, opened 2018-04-17
Alter face-element algorithm looping to ensure uniform face-element

alanw0:
ordinals in simd groups.

Our master-element and kernel implementations require that within
a simd group, all faces have the same element-ordinal.

Created a unit-test which alters a mesh to ensure at least two faces
with different element-ordinals are in the same bucket, then added
a custom kernel::execute method which checks for different ordinals
in the same simd group. Then put the corrected algorithm looping
in place to make the test pass.

### DIFF
--- a/include/AssembleFaceElemSolverAlgorithm.h
+++ b/include/AssembleFaceElemSolverAlgorithm.h
@@ -86,24 +86,38 @@ public:
 
         Kokkos::parallel_for(Kokkos::TeamThreadRange(team, simdBucketLen), [&](const size_t& bktIndex)
         {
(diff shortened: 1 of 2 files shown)


alanw0:
@spdomin I thought about the entity-sorter, but I think there is still a need for this looping change. I know the entity-sorter keeps faces with the same ordinal together, but does it put a bucket boundary between sets of different ordinals? If the boundary between different ordinals is within a bucket, then this looping change is still needed to be certain that the boundary doesn't occur within a simd group.

spdomin:
Right... Agreed... Feel free to process this pull request.

PR #368, opened 2018-04-23
Added shifted grad op for SCV

marchdf:
This is needed to enable a shifted grad op for the SCV terms like the production term for the RANS equations #366.

@spdomin and @rcknaus, how did you come up with the golds (e.g. `kokkos_me_gold::hex8_scv_grad_op`) for the unit tests? I would like to add unit tests for the SCV shifted grad ops before this gets merged but it's unclear to me how to do that. Do you have a script lying around that I could adapt?


rcknaus:
The gold grad_op values are just a comparison between the older Fortran routines and the new SIMD routines.  I don't have a script for all of the topologies, but you should be able to just do a patch-test with a linear field to check that they're accurate (pyramids will fail, unfortunately).

marchdf:
@spdomin and @rcknaus I added unit tests and did it similarly to what is currently done for SCV grad op. If you are ok with this, it is ready to merge.

PR #370, opened 2018-04-26
Create eigenvalue decomposition as its own class

jamelvin:
Extracted the eigenvalue decomposition from the AssembleScalarEigenEdgeSolverAlgorithm and created a separate EigenDecomposition class to handle this.  This will allow additional units, such as those involved with the new Hybrid model to have access to the EVD solver.

I also created unit tests for the EVD solvers.  

To test the implementation, I ran the ablNetural reg test and added the below to solution options to force it to invoke the EigenEdgeSolverAlgorithm and verified my solution was unchanged with the included code modifications.

eigenvalue_perturbation: yes

I didn't find any existing unit tests or reg tests that tested AssembleScalarEigenEdgeSolverAlgorithm, if there is one, please let me know and I will run it against that.  Thanks.

### DIFF
--- a/include/AssembleScalarEigenEdgeSolverAlgorithm.h
+++ b/include/AssembleScalarEigenEdgeSolverAlgorithm.h
@@ -24,7 +24,6 @@ namespace nalu{
 
 class Realm;
 template <typename T> class PecletFunction;
-class EigenDecomposition;
 
 class AssembleScalarEigenEdgeSolverAlgorithm : public SolverAlgorithm
 {
@@ -76,7 +75,6 @@ public:
 
   // peclect function specifics
   PecletFunction<double>* pecletFunction_;
-  EigenDecomposition eigSolver_;
 
   // constants and perturbation from user
   const double cGGDH_;
--- a/include/AssembleScalarEigenEdgeSolverAlgorithm.h
+++ b/include/AssembleScalarEigenEdgeSolverAlgorithm.h
@@ -11,6 +11,7 @@
 
 #include<SolverAlgorithm.h>
 #include<FieldTypeDef.h>
+#include<EigenDecomposition.h>
 
 namespace stk {
 namespace mesh {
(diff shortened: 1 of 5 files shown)


jamelvin:
Thanks for the recommendation @rcknaus, I have changed it to a namespace for now, although a class may be better depending on how best to handle the SIMD types.  For the SIMD types, should I

a) just overload the functions?
b) go back to a class and use templates? 

or were you thinking something else?  One of the other uses for these EVDs is in one of the new kernels we have created, which I had been using a temporary workaround on since when I first started exploring this form of the EVD solver for the metric tensor, I ran into some issues with the DoubleType and how to handle the exiting of the iteration loop and other if statement branching, since that seems to be incompatible with vectorization.  I don't know if you guys have any recommendations for that or example functions I could look at that do vectorization with similar situations? 

Is the best solution for this kind of stuff to just use the stk::math::if_then_else function and only exit the iteration loop when all of the vectorized entries satisfy the exit condition, i.e. checking the max/min over the vector?  Thanks.

 Adding @marchdf so he's in the loop.

rcknaus:
Both options are fine.  You can also use function templates.

As to SIMDizing the algorithm, what you suggested should work I think.  Otherwise, you can switch to the analytical version with a slightly loss of floating point accuracy.

jamelvin:
I have an update to this which includes support for simd data types and changes EigenDecomposition from a class to a namespace.  I will push these later today.

spdomin:
Dear @jamelvin, the np2 eigen teat failed last night after my merge. If you can resolve this is NaluCFD, that would be great. You may simply need to deactivate it in parallel. Best

jamelvin:
@spdomin I will look into that.  Sorry for the inconvenience.

jamelvin:
@spdomin It looks like I had a bug in my 2D solver for SIMD datatypes when handling diagonal matrices that caused the eigenvectors to be reversed 50% of the time.  I made a change to resolve this issue and it appears to work successfully now.  I am passing the unit tests in both serial and parallel running it many times.

What is the proper workflow here.  Do I just create a new pull request with the bug fix, or should I revert this one and then resubmit the whole thing?  Thanks.

spdomin:
@jamelvin, in this case, simply patching the bad 2D case, as you did, was great. I will let you know if my testing still shows an issue.

PR #373, opened 2018-05-01
Jroverf/face grad opt for tri and quad

overfelt:
WIP.  Just checking that this is what is needed for the Tri and Quad.  

This is just for the tri32D as an example of how to use the generic_grad_op_2d.  The quad should just be a call to the same function with a different traits object. 

I wrote the generic_grad_op_2d to replace the generic_grad_op_3d on the next commit.


src/master_element/Tri32DCVFEM.C
```diff
@@ -599,6 +599,21 @@ void Tri32DSCS::shifted_grad_op(
...
 //--------------------------------------------------------------------------
+void Tri32DSCS::face_grad_op(
+  int /*face_ordinal*/,
+  SharedMemView<DoubleType**>& coords,
+  SharedMemView<DoubleType***>& gradop)
+{
+  using traits = AlgTraitsEdge2DTri32D;
+
+  constexpr int derivSize = traits::numFaceIp_ * traits::nodesPerElement_ * traits::nDim_;
+  DoubleType psi[derivSize];
```
spdomin: Does this need an align?

overfelt: I can combine generic_grad_op_2d with generic_grad_op_3d so it is just generic_grad_op and works for all dimensions.  Then the same code will be used for all elements. 

Seems like when in doubt align is the thing to do.  I'll add it.

alanw0:
@overfelt @spdomin Note that there is also still a little bit of missing plumbing in ScratchViews also. Basically in each switch statement where there is currently a 'case SCS_FACE_GRAD_OP' you need to make sure there is a corresponding 'case SCS_SHIFTED_FACE_GRAD_OP'. And the view to be populated is dndx_shifted_fc_scs, corresponding to the non-shifted dndx_fc_scs.

include/master_element/MasterElementFunctions.h
```diff
@@ -118,15 +49,15 @@ namespace nalu {
...
+  void generic_grad_op(const GradViewType& referenceGradWeights, const CoordViewType& coords, OutputViewType& weights)
   {
     constexpr int dim = AlgTraits::nDim_;
 
     using ftype = typename CoordViewType::value_type;
     static_assert(std::is_same<ftype, typename GradViewType::value_type>::value,  "Incompatiable value type for views");
     static_assert(std::is_same<ftype, typename OutputViewType::value_type>::value,  "Incompatiable value type for views");
     static_assert(GradViewType::Rank   ==   3, "grad view assumed to be 3D");
-    static_assert(CoordViewType::Rank  == dim, "Coordinate view assumed to be spatial dimension");
+    static_assert(CoordViewType::Rank  ==   2, "Coordinate view assumed to be 2D");
```
spdomin: I do not understand this - sorry.. Can you please explain?

include/master_element/MasterElementFunctions.h
```diff
@@ -28,82 +28,82 @@
...
+
   template <typename AlgTraits, typename GradViewType, typename CoordViewType, typename OutputViewType>
-  void generic_grad_op_3d(const GradViewType& referenceGradWeights, const CoordViewType& coords, OutputViewType& weights)
+  void generic_grad_op(const GradViewType& referenceGradWeights, const CoordViewType& coords, OutputViewType& weights)
   {
+    constexpr int dim = AlgTraits::nDim_;
+
     using ftype = typename CoordViewType::value_type;
     static_assert(std::is_same<ftype, typename GradViewType::value_type>::value,  "Incompatiable value type for views");
     static_assert(std::is_same<ftype, typename OutputViewType::value_type>::value,  "Incompatiable value type for views");
```
overfelt: Asserts on numeric type:  All double or all FAD or all SIMD.  I think this would be a problem if the coordinates were double and the weights were SIMD.  The code would compile since it is valid to multiply a SIMD by a double, but it would be incorrect as we know that a SIMD weight would be applied only to a SIMD coordinate.

include/master_element/MasterElementFunctions.h
```diff
@@ -28,82 +28,82 @@
...
-    static_assert(GradViewType::Rank == 3, "grad view assumed to be 3D");
-    static_assert(CoordViewType::Rank == 2, "Coordinate view assumed to be 2D");
-    static_assert(OutputViewType::Rank == 3, "Weight view assumed to be 3D");
-    static_assert(AlgTraits::nDim_ == 3, "3D method");
+    static_assert(GradViewType::Rank   ==   3, "grad view assumed to be 3D");
+    static_assert(CoordViewType::Rank  ==   2, "Coordinate view assumed to be 2D");
+    static_assert(OutputViewType::Rank ==   3, "Weight view assumed to be 3D");
```
overfelt: A bunch of rank asserts.  Would probably not compile if the wrong rank object were passed anyway but this will give a clearer compile error instead of something like an incompatible data types error in the body of the function.

include/master_element/MasterElementFunctions.h
```diff
@@ -28,82 +28,82 @@
...
+    static_assert(AlgTraits::nDim_     == dim, "2D method");
```
overfelt: This is just a bad assert since dim is set to AlgTraits::nDim_ to begin with no need to assert they are equal.

PR #374, opened 2018-05-02
Towards the penalty approach for element-based open.

spdomin:
Design-order penalty continuity approach has been established. Time to deprecate
the old (Pbip - Pscs_op) approach.

* This push required a number of plumbing fixes to the former face/element
approach. BC creation is currently automatic. Formerly, symmetry bc
was being created if the lumped_mass and adveciton element source terms
were active. I fixed this to use automatic. Diffs expected as now only
one symmetry kernel will be called:)

* Generalized the face/element kernel creation allowing for
quad4/hex8/pyr5/wedge6 and tri3/tet4/pyr5/wedge6 combinations. Before,
it was only creating quad4/hex8 and tri3/tet4. This covers the full
hybrid element topos.

* Provide elemOrdinal to all face/elem kernels through a new argument. This
resolves the former FIXME #2 in which face_ordinal was hard coded. This also
means that the former symmetry run was possibly using the wrong ordinals.

Now, proceed with the new Continuity consolidated bc.

* Added ContinuityOpenElemKernel

* Added ComputeMdotElemOpenPenaltyAlg (will drop the penalty once we deprecate
the former capability). Now, if a user asksfor a consolidated approach,
the new path forward is provided.

* Hooked in the new open penalty approach for ContEQS for this consolidated
bc approach.

Notes:

a) I crossed implementations here as I started on the penalty approach before
I realized claen-up of face/elements was required. Deactivate penalty
appraoch while I sort through the possible issue.
b) symmetry seems to work, however, the model case is uniform flow..

### DIFF
--- a/include/kernel/ContinuityOpenElemKernel.h
+++ b/include/kernel/ContinuityOpenElemKernel.h
@@ -5,13 +5,8 @@
 /*  directory structure                                                   */
 /*------------------------------------------------------------------------*/
 
-#ifndef ScalarOpenAdvElemKernel_h
-#define ScalarOpenAdvElemKernel_h
-
-#include "master_element/MasterElement.h"
-
-// scratch space
-#include "ScratchViews.h"
+#ifndef ContinuityOpenElemKernel_h
+#define ContinuityOpenElemKernel_h
 
 #include "kernel/Kernel.h"
 #include "FieldTypeDef.h"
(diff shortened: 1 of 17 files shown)


include/kernel/ContinuityOpenElemKernel.h
```diff
@@ -0,0 +1,83 @@
...
+  ScalarFieldType *pressureBc_{nullptr};
+  ScalarFieldType *density_{nullptr};
+  GenericFieldType *exposedAreaVec_{nullptr};
+
+  double projTimeScale_{1.0};
+  const double mdotCorrection_{0.0};
+  const double penaltyFac_{2.0};
+  
+  const bool shiftedGradOp_;
+  const bool reducedSensitivities_;
```
alanw0: This is a minor concern, maybe even stylistic, but I like to see all the class members either be initialized here, or in the constructor initializer list, not a mixture.

spdomin: I think I followed your suggestion the last time you reviewed something in that I pushed pure connate to the header and things that are user-specified, however, constant, to the constructor.

PR #378, opened 2018-05-14
FaceElem alg should use nodesPerElem not nodesPerFace

spdomin:
* Removed incorrect usage of nodesPerFace_ for apply_coeff and connected_nodes.

* new gold for elemHybrid

Notes:

a) advection-only bcs looks good now (mixture fraction), however, open
continuity still fails the first couple of steps in which the new consolidated
bc approach is activated.
b) diffs expected. Symmetry still likely wrong - especially if there is
something off in exposed grad-op.


alanw0:
Looks fine to me.

PR #389, opened 2018-05-22
Fix aligned declarations for SIMD data types

sayerhs:
- Define NALU_ALIGNED that uses KOKKOS_MEMORY_ALIGNMENT instead of the hard-coded 64 used in the code currently.

- Define new type-alias AlignedView that enforces Kokkos::MemoryTraits<Kokkos::Aligned>

- Ensure all declarations of DoubleType arrays in Kernels and MasterElement sources are aligned types.

### DIFF
--- a/include/KokkosInterface.h
+++ b/include/KokkosInterface.h
@@ -12,7 +12,7 @@
 #include <Kokkos_Macros.hpp>
 #include <Kokkos_Core.hpp>
 
-#define NALU_ALIGN(size) __attribute__((aligned(size)))
+#define NALU_ALIGNED alignas(KOKKOS_MEMORY_ALIGNMENT)
 
 #if defined(__INTEL_COMPILER)
 #define POINTER_RESTRICT restrict
@@ -34,6 +34,9 @@ using TeamHandleType = Kokkos::TeamPolicy<DeviceSpace, DynamicScheduleType>::mem
 
 template <typename T>
 using SharedMemView = Kokkos::View<T, Kokkos::LayoutRight, DeviceShmem, Kokkos::MemoryUnmanaged>;
+
+template<typename T>
+using AlignedViewType = Kokkos::View<T, Kokkos::MemoryTraits<Kokkos::Aligned>>;
 
 using DeviceTeamPolicy = Kokkos::TeamPolicy<DeviceSpace>;
 using DeviceTeam = DeviceTeamPolicy::member_type;
--- a/include/SimdInterface.h
+++ b/include/SimdInterface.h
@@ -16,6 +16,7 @@
  */
 
 #include "stk_simd/Simd.hpp"
+#include "Kokkos_Macros.hpp"
 
 #include <vector>
 
@@ -27,7 +28,7 @@ typedef stk::simd::Double SimdDouble;
 typedef SimdDouble DoubleType;
 
 template<typename T>
-using AlignedVector = std::vector<T, non_std::AlignedAllocator<T, 64>>;
+using AlignedVector = std::vector<T, non_std::AlignedAllocator<T, KOKKOS_MEMORY_ALIGNMENT>>;
 
 using ScalarAlignedVector = AlignedVector<DoubleType>;
 
--- a/include/kernel/ContinuityAdvElemKernel.h
+++ b/include/kernel/ContinuityAdvElemKernel.h
@@ -69,7 +69,7 @@ private:
   const double om_interpTogether_;
 
   // scratch space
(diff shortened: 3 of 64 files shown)


sayerhs:
@spdomin @rcknaus Here is a possible implementation of aligned DoubleType whenever SIMD types are used. I could not find a consistent way to get a type-alias that would work on Intel, GCC and Clang compilers. So it was not possible to shorted `NALU_ALIGNED DoubleType` to `AlignedDoubleType`. 

One question for you is: should we also replace all [such declarations](https://github.com/NaluCFD/Nalu/blob/master/src/kernel/ContinuityMassElemKernel.C#L91) with aligned declarations?

sayerhs:
One remaining issue is the warning related to over-aligned new with GCC v8.x in PecletFunction that @rcknaus reported. I am not sure what is the best way to fix that, we can either pass `-faligned-new=64` to the compiler, or provide a custom new operator for the PecletFunction. Any thoughts on this?

sayerhs:
@spdomin I think you started reviewing before I pushed commit ed23074. I have updated all `const ftype` declarations to also use `NALU_ALIGNED`. 

I look forward to comments/suggestions from @rcknaus and @alanw0 on enforcing alignment on all `DoubleType` declarations (including non-array ones).

overfelt:
So, was it not possible to use alignas() to force alignment?  Or is that still an option....

overfelt:
What part of 
`using AlignedDoubleType alignas(KOKKOS_MEMORY_ALIGNMENT) = DoubleType;`
did not work?  I tried this in a trivial test program and it seemed to align correctly PODs and simple structs.  I did not try with Sacado types.

sayerhs:
@overfelt That threw warnings when I attempted to do 

```
AlignedDoubleType someArray[AlgTraits::nDim_];
```

If this can be made to work consistently, then `AlignedDoubleType` would be my preferred approach.

sayerhs:
@overfelt Here is the sample code and interactions with compilers 

```
#include <iostream>

int main()
{
    using AlignedDoubleType alignas(alignof(double)) = double;

    AlignedDoubleType myArray[3];

   std::cout << "Success!!" << std::endl;

    return 0;
}

---------------------------------------------------------------------
$ g++ --version; g++ -std=c++11 test.cpp && ./a.out
g++ (GCC) 6.2.0
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

Success!!

---------------------------------------------------------------------
$ icpc --version; icpc -std=c++11 test.cpp && ./a.out
icpc (ICC) 18.0.1 20171018
Copyright (C) 1985-2017 Intel Corporation.  All rights reserved.

test.cpp(5): error: identifier "AlignedDoubleType" is undefined
      using AlignedDoubleType alignas(alignof(double)) = double;
            ^

test.cpp(5): error: expected a ";"
      using AlignedDoubleType alignas(alignof(double)) = double;
                              ^

test.cpp(7): error: identifier "AlignedDoubleType" is undefined
      AlignedDoubleType myArray[3];
      ^

compilation aborted for test.cpp (code 2)
```

sayerhs:
Alternately, if you don't use `alignof(double)` then this is the warning from `g++` 

```
test.cpp: In function ‘int main()’:
test.cpp:8:32: error: alignment of array elements is greater than element size
     AlignedDoubleType myArray[3];
                                ^
```

sayerhs:
@spdomin @overfelt @rcknaus I have updated the pull request based on our discussions yesterday and this pull request is ready to be merged pending your review. 

Summary of changes:

- Define new macro `NALU_ALIGNED` that uses C++ `alignas` keyword and `KOKKOS_MEMORY_ALIGNMENT` that replaces `NALU_ALIGN(64)` 

- New typedef `AlignedViewType` that enforces `Kokkos::MemoryTraits::Aligned` on `Kokkos::View`s used in Kernels. 

- Replace `NALU_ALIGN(64)` with `NALU_ALIGNED` everywhere, replace `Kokkos::View<...>` with `AlignedViewType<...>` everywhere. 

- Update all `master_element` and `kernel` source code to enforce alignment of array declarations.

Some comments:

- I dropped the idea of `#define NALU_ALIGNED_DOUBLE alignas(KOKKOS_MEMORY_ALIGNMENT) DoubleType` because we would still need `NALU_ALIGNED` for the `ftype` declarations in `MasterElementFunctions.h` 

- The usage is `NALU_ALIGNED DoubleType` and not `DoubleType NALU_ALIGNED` because of the following warning issued in `gcc-6.2.0`

```
test.cpp: In function ‘int main()’:
test.cpp:12:11: warning: attribute ignored [-Wattributes]
    double alignas(64) myArray1[3];
           ^~~~~~~
```

PR #394, opened 2018-05-31
Change Pyr5 SCV and SCS to support design-order Pyramid runs.

spdomin:
* Modified SCS quad points (8ip -> 12ip, new location)

* Modified SCV quad points (same numIp, new location)

* added new bent hex approach for non-consolidated SCV

* added new bent hex approach for consolidated Kokkos SCV

* lrscv added.

* fixed bug in edge-area vector assembly. We can not loop over
edges since pyramids have two ips per face in some instances.
This drove the addition of a new ME data structure to map
the given ip to the proper edge. Make sure that the ip is
used rather than nedge! However, with these changes, edge-based
is now clean with only very slights diffs due to slighlty different
volume and edge area vec at edge pyramid interfaces.

* added shifted_pyr_deriv and shifted_pyr_derivative.

* shifted now uses old basis due to an instability found when
using the new basis in a shifted config.

* changed PyrSCS::face_grad_op() to now always use the shifted
for tri3 faces since this reverts to the old approach.

Notes:

a) The new basis and quadrature for the Pyr5 has been shown to
be design-order accurate. However, when shifting the Pyr5 grad_op,
instability arises. This drove the support of both basis definitions
as we need to support a shifted operator.

b) I have run just about every elem hybrid case with and without shifting
and mix/matching consolidated and non-consolidated and all match. As such,
I have high confidence in this large project.

c) As with b, I have also run a large number of edge-based hybrid runs and
they, again, all look great and stable.


spdomin:
@mbarone81, thank you for the notes. A redesign would be useful. For that, I will pull in @overfelt and/or @rcknaus. Personally, I like a design that calls the shifted directly as in other examples. Making this change would percolate up to the ElemData level.

Will capture the technical nuance of two basis definitions in the theory manual.

spdomin:
@jrood-nrel, feel free to pull this change into Nalu-wind.

This should be the last sync that you (personally) will need to process.

PR #395, opened 2018-06-04
Nalu changes to utilize Mark Hoemmen's recent Tpetra optimizations

alanw0:
1. Use parameter "compute triangular constants" -> false
To avoid the expense of determining whether the matrix graph is triangular
and how many diagonals it has on each processor.

2. Use a new Importer constructor which accepts an array telling it which
processor each remote column index comes from. This way Tpetra doesn't have
to communicate to obtain that information itself.

### DIFF
--- a/include/TpetraLinearSystem.h
+++ b/include/TpetraLinearSystem.h
@@ -15,7 +15,6 @@
 
 #include <Tpetra_DefaultPlatform.hpp>
 #include <Kokkos_DefaultNode.hpp>
-#include <Kokkos_UnorderedMap.hpp>
 #include <Tpetra_Vector.hpp>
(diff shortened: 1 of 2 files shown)


spdomin:
Looks good. Many thanks.

spdomin:
@jrood-nrel, feel free to pull in this PR to Nalu-Wind. The MPI call in question was already part of the previous Tpetra optimization. However, this code change does, in fact, require an update to Trilinos.

spdomin:
Actually, @alanw0 Nalu tests differ last night. As such, @jrood-nrel should hold off - if he does chose to move forward with the merge. I am looking into it, however, looks clear that Sunday night passed while Monday failed.

PR #396, opened 2018-06-06
Clean up SIMD face_grad_op for pyr/wedge

rcknaus:
* The masking operation was unnecessary after the switch to using uniform collection of side ordinals.  Additionally, the wedge had an error where it was accessing some uninitialized values.


spdomin:
Looks good. This cleans up our verification tests.

PR #443, opened 2019-04-10
Update Hypre to work at SNL with Nalu

spdomin:
* Patch required Hypre changes based on linSys interface change

* Re-bless nc test

Notes:

a) Ran Hypre and Muelu for dcnc test for 1000 time steps. Results matched,
   H solve count lower than M, however, M faster by ~10%.


spdomin:
Many thanks to @sayerhs for his work on Hypre in Nalu.

PR #459, opened 2019-08-09
Fix missing gkkSq term in WALE

spdomin:
* fix the theory source

* re-bless rtests and unit tests due to wale tvisc change


src/TurbViscWaleAlgorithm.C
```diff
@@ -88,8 +91,12 @@ TurbViscWaleAlgorithm::execute()
     for ( stk::mesh::Bucket::size_type k = 0 ; k < length ; ++k ) {
 
       const double *dudx = stk::mesh::field_data(*dudx_, b[k] );
+      
+      double gkkSq = 0.0;
+      for ( int i = 0; i < nDim; ++i ) {
+        gkkSq += dudx[nDim*i]*dudx[nDim*i];
```
rcknaus: There should be more mixed terms from squaring the matrix here.  In 2D, the trace term is dudx^2 + dvdy^2 + 2 dudy dvdx .

spdomin: Right. I will add that.

PR #493, opened 2020-02-05
Fix simd usage to not access the '.data' member's operator[]

alanw0:
Coming changes to stk-simd will use a different type for the
._data member, which won't have operator[].

stk-simd is going to use Dan Ibanez' kokkos-simd prototype
as a backend. That prototype mirrors a proposed simd addition
to the C++23 standard. Additionally, it adds support for a
couple of instruction-sets that stk-simd doesn't currently
support, namely ARM and IBM/POWER.


spdomin:
Sorry, I missed this. However, this clarifies our side conversation:)

Many thanks..

PR #496, opened 2020-02-27
clean up non-interacting periodic and non-conformal alg

spdomin:
* dof STATUS is still a mess. Very fragile and in great need of a
  redisn. This push at least allows for non-interacting dg and periodic
  bcs to be properly handled... The previous implementation was confusing
  the dof status for a non conformal scheme when that node was not a
  perioidc node, however, periodicity was activated.

Zero diffs; several use cases testng this all work now.

### DIFF
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -2179,37 +2179,39 @@ int getDofStatus_impl(stk::mesh::Entity node, const Realm& realm)
   const bool entityIsShared = b.shared();
   const bool entityIsGhosted = !entityIsOwned && !entityIsShared;
 
-  bool has_non_matching_boundary_face_alg = realm.has_non_matching_boundary_face_alg();
-  bool hasPeriodic = realm.hasPeriodic_;
-
-  if (realm.hasPeriodic_ && realm.has_non_matching_boundary_face_alg()) {
-    has_non_matching_boundary_face_alg = false;
-    hasPeriodic = false;
-
-    stk::mesh::Selector perSel = stk::mesh::selectUnion(realm.allPeriodicInteractingParts_);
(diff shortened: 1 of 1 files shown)


src/TpetraLinearSystem.C
```diff
@@ -2218,33 +2220,24 @@ int getDofStatus_impl(stk::mesh::Entity node, const Realm& realm)
       return DS_SharedNotOwnedDOF;
```
alanw0: I'm confused by "if owned return shared". Looks like this already existed before your commit.

src/TpetraLinearSystem.C
```diff
@@ -2271,7 +2264,16 @@ int getDofStatus_impl(stk::mesh::Entity node, const Realm& realm)
       }
     }
   }
-
+  
+  // finally, if non conformal is active
+  if ( nonConformalActive ) {
+    if (entityIsOwned)
+      return DS_OwnedDOF;
+    if (!entityIsOwned && (entityIsGhosted || entityIsShared)){
```
alanw0: I'm still not sure this is correct either...

spdomin: This is why dofStatus is a mess. In STK, we have clean definitions of owned, shared and ghosted.  However, dofStatus does not respect this. For the non-conformal all to work properly, a ghosted node due to the non-conformal all needs to be listed as "sharedNotOwned". Every part of the code in Tpetra assumes this is the case.

Sadly, in that symmetry bc example, the ghosted node due to non-conformal was "correct" in listing it as "shared not owned".

I guess that in dofStatus, we have taken over the definition of shared. It means not owned, however, ghosted or shared when the non-conformal all is activated.

spdomin:
agreed... This is just a bandage for this periodic/nc-alg use case.

Issue #15, opened 2016-10-13
Surface force and moment post processing zeros out fields when not applying


spdomin:
In my LES channel flow, I am computing a time mean tau_wall, Yplus, etc. However, the algorithm is only executed at a user-defined output step. Meanwhile, the field is always zeroed each and every time step. 

Bottom line: the instantaneous tau_wall is fine, however, the time mean is not correct unless the algorithm is applied each and every time step.

Fix: In the AlgDriver, make sure that we only zero, execute, etc., if the frequency is met.

@mbarone81 this is just a FYI as I plan on fixing this today.

spdomin:
Resolved by https://github.com/NaluCFD/Nalu/pull/498

Thanks to Me469 for the motivation.

PR #498, opened 2020-03-05
Attempt at resolving averaging issue

spdomin:
Not ready for prime time... Incremental


spdomin:
Looks like all is well with this implementation.

Consider removing frequency per block specification as this complicates the internal logic.

https://github.com/NaluCFD/Nalu/issues/15

Issue #500, opened 2020-04-06
Slow-down in sliding mesh sims


spdomin:
dgThreeBlade looks much slower now. This was seemingly due to the fix for periodic and sliding mesh being run in the same run:

https://github.com/NaluCFD/Nalu/commit/f9ea67b5fe4dba99cb0035ea6d0babc68a792d7a

Basically, around the 27th or 28th of February, 2020.

spdomin:
@alanw0, could you please look at this Coe change and let me know if you see anything silly? My Mac runs seem ~2x slower now that this change is in the code base. Linux also showed a substantial slow down.

spdomin:
Yes... Note in my slow-down, periodic is not active. Does the code still loop over all nodes? I guess I can check.

I am surprised that this change caused such a slowdown. However, the data seems clear.

spdomin:
@alanw0, is:

if (entityIsShared && !entityIsOwned)

the same as:

if ( entityIsShared ) 

In other words, if a node is owned by P0, is that node on P0 also shared? I think not..

spdomin:
I have a code implementation that only performs the detailed check over all nodes if both periodic and nonconformal are active. This resolved the slow down. Once I push and test this branch (I am testing on my Mac), I will close this issue.

spdomin:
ctest -R dgNonConformalThreeBlade
Test project /Users/naluIt/gitHubWork/nightlyBuildAndTest/Nalu/build
    Start 14: dgNonConformalThreeBlade
1/1 Test #14: dgNonConformalThreeBlade .........   Passed  103.38 sec

No longer 300-600 seconds...

alanw0:
Any combination of shared and owned is possible for an entity. i.e., since shared and owned are logically booleans, they represent 4 possible states for an entity, and all are possible.

spdomin:
https://github.com/NaluCFD/Nalu/pull/501

PR #501, opened 2020-04-07
Provide a detailed check on periodic and ncAlg node if both are active

spdomin:
Attempts to resolve slowdown noted in the following issue:

https://github.com/NaluCFD/Nalu/issues/500

However, I need to test the full set of cases on a linux box.


spdomin:
All pass with time back to 100 seconds.

PR #502, opened 2020-04-25
fixed two typos in documentation

mulligatawny:


spdomin:
Thanks. We plan to deprecate this capability and point users to Nalu-Wind.

PR #527, opened 2021-02-23
6 DOF motion w/ testing

wjhorne:
Includes the following functionality

- 6 DOF motion of arbitrary blocks according to body coordinates using Euler rigid body dynamics
- Applied const forcing via input file
- Forces/Moments applied to movement of mesh blocks using integrated forces on selected surfaces
- Arbitrary block mesh velocity/displacement both rectilinear and angular

New regression test added that illustrates setup and use for an example overset case of a sphere freely falling in a quiescent box.


spdomin (approved):
This looks good for a first step. Much of the clean-up will be to abstract out different types of mesh motion. I also believe that we need to remove the specific calculation of forces and moments in favor of using wall friction velocity.

spdomin:
I am not sure why previous SuperLU work showed up here. Perhaps at your next pull request, you can have a new branched code base.

spdomin (commented):
Simple mistake:) Good catch.

PR #550, opened 2021-08-05
dpdx fringe update

wjhorne:
This change fixes a potential issue with dp/dx when simulating moving mesh cases when using overset. dp/dx is NaN inside of removed regions while using overset. These NaN values can appear in the solution when a removed cell emerges into the solution requiring a fringe reconstruction.


spdomin (approved):
This looks good. You may want to determine if the projected nodal gradients are required everywhere as well? For example, we use dudx, dvofdx, etc. and these are not currently included in the register_overset_bc() logic.

Also, convert the vofFallingSphere when you have a chance to the buoy case.

Thanks!

Please look into this. In the meantime, I will approve since this is a required change.

PR #555, opened 2021-09-29
ioBroker signature changes

spdomin:


spdomin:
@alanw0 - we should be good to go on the Nalu side.

alanw0:
> @alanw0 - we should be good to go on the Nalu side.

Great, thanks.

PR #557, opened 2021-11-11
NaN Overset Fringe Updates

wjhorne:
This fixes first time step CFL restrictions that were previously producing immediate NaN values depending on the initial CFL. This was affecting cases in other ways as indicated by the differences in the .norm files in our overset tests.


spdomin (approved):
Looks good and as we discussed.

PR #559, opened 2021-11-18
Buoyancy VOF stabilization change

wjhorne:
This moves buoyancy over to the VOF pressure stabilization. There are some substantive changes in VOF solutions due to this change as evidenced by the changes in all of the gold files for VOF tests. So far I am seeing lessened overset perturbations with the changes. Tests/cases outside of the regression need to still be run to check the impact. This would include damBreak and more.


reg_tests/test_files/vofSlosh/vofSlosh.norm.gold
```diff
@@ -1,200 +1,200 @@
-3.121984735997264 1 0.001
```
wjhorne, line 1: I was really struck by how much this initial residual changed between here and the new stabilized result.

spdomin:
@wjhorne, I will start testing this branch on my suite of cases and we can discuss the overall behavior.

spdomin (commented):
Looks clean. I will pull and test tomorrow. 

After we have the results of a few cases, we can push...

spdomin (changes requested):
Wait, does this mean that vofBuoy is back to the original state? After all, the buoyancy in the stabilization is now off - back to the original.

wjhorne:
The behavior I was going for is that when the balanced force approach is activated the buoyancy stabilization is activated by default. If there is no balanced force the stabilization is deactivated by default. The defaults are ignored if a user turns on or off the buoyancy stabilization in the input. The reasoning is that the buoyancy stabilization appears in the VoF routines related to the balanced force approach, but not elsewhere. I'll take a quick look here to make sure that I didn't do something silly with the latest commit to overwrite what I intended.

wjhorne:
vofBuoy.i looks correct from what I am seeing under the files changed. It should activate the new stabilization and then turn off the buoyancy term since it is included in the new stabilization through the pressure gradient and not through an explicit buoyancy term in momentum.

spdomin (commented):
Let's talk in person about this set of nuanced logic. If BF is not active, then the buoyancy pressure stabilization is not germane. I would simply default buoyancyPressureStab_ to true, balancedForce_ to false, and then call a single get_if_present(). If we want to be really careful, we can have a "look ahead" on all of these line commands along with source terms in momentum to make sure that we don't allow someone to turn on a buoyancy momentum source if the buoyancy pStab is active.

spdomin (approved):

spdomin:
Error Checks To do: 

1. clean up logic to check for double counting.
2. Ensure that new stand is not allowed when balanced force is not allowed.

PR #566, opened 2021-12-07
Revert "Buoyancy VOF stabilization change"

spdomin:
Reverts NaluCFD/Nalu#559


spdomin:
Found a few issues that need to be resolved.

PR #567, opened 2021-12-08
Buoyancy stabilization

wjhorne:
Removed buoyancyWeight inputs to Algorithm and kernel constructors and fixed issues related to accidental 6-DOF change. 

Diffs in vofBuoy are mostly due to fixes to FixedHeight user function with some differences due to stabilization. 
Diffs in generator case are due to added stabilization.


src/VolumeOfFluidEquationSystem.C
```diff
@@ -117,7 +117,8 @@ VolumeOfFluidEquationSystem::VolumeOfFluidEquationSystem(
     smooth_(smooth),
     smoothIter_(smoothIter),
     isInit_(true),
-    scsAdvection_(false)
+    scsAdvection_(false),
+    buoyancyStab_(realm_.solutionOptions_->buoyancyPressureStab_ ? 1.0 : 0.0)
```
spdomin, line 121: I don't think you ever use this... As such, we should remove it solely in favor of using solutionOptions member data. Let me know if I am wrong.

wjhorne: I believe this is leftover from when I was feeding in the value into the individual kernels/algorithms. It should be removed without issue. I'll do so here in a moment and make sure it all still works

reg_tests/test_files/vofWaveGenerator/vofWaveGenerator.i
```diff
@@ -237,11 +237,12 @@ realms:
       use_consolidated_face_elem_bc_algorithm: yes
 
       activate_balanced_force_algorithm: yes
+      activate_buoyancy_pressure_stabilization: yes
```
spdomin, line 240: This was to be perfectly explicit, as when BF is active, buoyancy pressure stabilization is also (similar to vofBuoy)... Keep this as I like the inclusion of this line command.

spdomin (changes requested):
I think there is one last iteration - specifically removal of the buoyancy stabilization member data in VofEQS. This is never required based on the reliance to solution options.

In reality, I think you could have simply created a buoyancy weight that you pass around... Either way is fine with me.

src/VolumeOfFluidEquationSystem.C
wjhorne: This has now been removed in latest push.

spdomin (approved):

PR #570, opened 2021-12-15
Remove aggregation: phase2

spdomin:
* change in Muelu XML. phase2 line command is gine (causes a throw) in favor
  of: "aggregation: match ML phase2a". However, this line command is
  false.

* remove old line command. Defaults in Muelu are such that match ML phase2a is false.

Notes:

a) no diffs [expected]


spdomin:
As per: https://github.com/trilinos/Trilinos/pull/9747#event-5774565762

Thanks @jhux2 and Christian.

PR #571, opened 2022-01-14
Re-bless recent Trilinos BLAS related changes

wjhorne:
Updates gold files to accept recent changes in Trilinos related to BLAS routines.


spdomin:
Feel free to close the Trilinos ticket after the nightly test suite passes. If I see this info first, I can handle that.

PR #575, opened 2022-01-25
Move Tpetra over to Kokkos Views

wjhorne:
Fixes host of deprecated code in Tpetra routines through the use of Kokkos::Views rather than Teuchos. Type definitions in Tpetra are used rather than directly creating Kokkos::Views to ensure that Device/Host memory layouts, etc are correct no matter the system.

### DIFF
--- a/src/TpetraLinearSystem.C
+++ b/src/TpetraLinearSystem.C
@@ -1153,7 +1153,12 @@ void verify_row_lengths(const LinSys::Graph& graph,
          <<" entries: ";
       GlobalOrdinal rowGID = graph.getRowMap()->getGlobalElement(i);
       std::vector<GlobalOrdinal> vIndices(graph.getNumEntriesInGlobalRow(rowGID));
-      Teuchos::ArrayView<GlobalOrdinal> colIndices(vIndices);
+
+      Tpetra::CrsGraph<>::nonconst_global_inds_host_view_type colIndices("col_indices", vIndices.size());
+      for (size_t col_idx = 0; col_idx < vIndices.size(); ++col_idx) 
+        colIndices(col_idx) = vIndices[col_idx];
+      
+
       size_t rowLen = 0;
       graph.getGlobalRowCopy(rowGID, colIndices, rowLen);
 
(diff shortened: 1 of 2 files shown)


spdomin (approved):
Looks good. @alanw0 could you provide your input so that we can close the PR? Many thanks.

src/TpetraLinearSystem.C
```diff
@@ -1153,7 +1153,12 @@ void verify_row_lengths(const LinSys::Graph& graph,
          <<" entries: ";
       GlobalOrdinal rowGID = graph.getRowMap()->getGlobalElement(i);
       std::vector<GlobalOrdinal> vIndices(graph.getNumEntriesInGlobalRow(rowGID));
-      Teuchos::ArrayView<GlobalOrdinal> colIndices(vIndices);
+
+      Tpetra::CrsGraph<>::nonconst_global_inds_host_view_type colIndices("col_indices", vIndices.size());
+      for (size_t col_idx = 0; col_idx < vIndices.size(); ++col_idx) 
+        colIndices(col_idx) = vIndices[col_idx];
```
alanw0, line 1159: I think this loop is unnecessary since vIndices wasn't initialized/filled yet, so no need to copy its contents.

src/TpetraLinearSystem.C
```diff
@@ -1174,7 +1179,9 @@ void dump_graph(const std::string& name, int counter, int proc, LinSys::Graph& g
   for(size_t i=0; i<myGlobalIndices.size(); ++i) {
     GlobalOrdinal rowGID = myGlobalIndices[i];
     std::vector<GlobalOrdinal> vIndices(graph.getNumEntriesInGlobalRow(rowGID));
-    Teuchos::ArrayView<GlobalOrdinal> colIndices(vIndices);
+    Tpetra::CrsGraph<>::nonconst_global_inds_host_view_type colIndices("col_indices", vIndices.size());
+    for (size_t col_idx = 0; col_idx < vIndices.size(); ++col_idx) 
+      colIndices(col_idx) = vIndices[col_idx];
```
alanw0, line 1184: Same here, no need to loop a copy contents of vIndices. colIndices will be filled by the following call to graph.getGlobalRowCopy.

alanw0 (approved):
Looks pretty good for the most part. I made a couple comments about loops that I think are not needed.

wjhorne:
I'll remove the loops and add it to the PR this afternoon

src/TpetraLinearSystem.C
spdomin: Is there any sensitivity over this name, "col_indices"? We may have multiple equation systems sitting around and I wonder if this name needs to be specialized to the linear system? In general, most systems are the same size (1x1) with exception of the 3x3 momentum system. My guess is that the regression tests would have picked something up.

spdomin (approved):
Looks great.

PR #587, opened 2022-05-18
Fix Trilinos deprecations

wjhorne:
This fixes the latest Trilinos deprecation warnings. As far as I can tell this is all that is left. There are still some deprecation warnings within Trilinos itself that hopefully will get sorted soon.


spdomin:
Looks clean.

PR #593, opened 2022-10-14
Updated Nalu for the new STK Mesh simple_fields workflow

djglaze:
STK Mesh is in the process of simplifying how Fields are registered and managed, to remove a perpetual source of user errors, inconsistency, and confusion.  The change essentially involves removing the extra template parameters beyond the datatype from Fields.  For example, setting up and using a 3-component vector Field used to look like this:

```
  using VectorField = stk::mesh::Field<double, stk::mesh::Cartesian3d>;
  VectorField & field = meta.declare_field<VectorField>(stk::topology::NODE_RANK, "velocity");
  stk::mesh::put_field_on_mesh(field, meta.universal_part(), 3);

  // ... (later retrieval for usage)
  VectorField & field = *meta.get_field<VectorField>(stk::topology::NODE_RANK, "velocity");
```

Now, it looks like this:

```
  using VectorField = stk::mesh::Field<double>;
  VectorField & field = meta.declare_field<double>(stk::topology::NODE_RANK, "velocity");
  stk::mesh::put_field_on_mesh(field, meta.universal_part(), 3);
  stk::io::set_field_output_type(field, "Vector_3D");

  // ... (later retrieval for usage)
  VectorField & field = *meta.get_field<double>(stk::topology::NODE_RANK, "velocity");
```

Notable differences:
  - Fields are now *only* templated on their datatype
  - Sizing is done *only* in the `put_field_on_mesh()` call.  (This was optional and redundant before.)
  - `declare_field()` and `get_field()` functions are *only* templated on the datatype of the Field and not the Field itself.
  - If you care about variable component subscripting in an output Exodus file, call `set_field_output_type()`.  The above example will subscript the Field components as `[_x, _y, _z]`.  If this is omitted, the subscripts will default to `[_1, _2, _3]`.  Scalar Fields are unaffected.

As a migration aid, `MetaData::use_simple_fields()` is called early in construction.  This changes any mistaken use of old-style calls to a run-time error, until the old APIs are eventually deprecated and removed.

### DIFF
--- a/docs/source/theory/codeAbstractions.rst
+++ b/docs/source/theory/codeAbstractions.rst
@@ -261,12 +261,11 @@ elemental, nodal, face and edge of desired size.
       const int numStates = realm_.number_of_states();
 
       // declare it
-      density_ 
-        =  &(metaData_.declare_field<ScalarFieldType>(stk::topology::NODE_RANK, 
-                                                     "density", numStates));
+      density_ = &(metaData_.declare_field<double>(stk::topology::NODE_RANK, 
+                                                  "density", numStates));
 
       // put it on this part
-      stk::mesh::put_field(*density_, *part);
+      stk::mesh::put_field_on_mesh(*density_, *part, nullptr);
     }
 
 :math:`\bullet` edge field registration:
@@ -278,10 +277,10 @@ elemental, nodal, face and edge of desired size.
       stk::mesh::Part *part)
     {
       const int nDim = metaData_.spatial_dimension();
-      edgeAreaVec_ 
-        = &(metaData_.declare_field<VectorFieldType>(stk::topology::EDGE_RANK, 
-                                                    "edge_area_vector"));
-      stk::mesh::put_field(*edgeAreaVec_, *part, nDim);
+      edgeAreaVec_ = &(metaData_.declare_field<double>(stk::topology::EDGE_RANK, 
+                                                      "edge_area_vector"));
+      stk::mesh::put_field_on_mesh(*edgeAreaVec_, *part, nDim, nullptr);
+      stk::io::set_field_output_type(*edgeAreaVec_, stk::io::FieldOutputType::VECTOR_3D);
     }
 
 :math:`\bullet` side/face field registration:
@@ -299,9 +298,9 @@ elemental, nodal, face and edge of desired size.
         stk::topology::rank_t sideRank 
           = static_cast<stk::topology::rank_t>(metaData_.side_rank());
         GenericFieldType *wallFrictionVelocityBip 
-          =  &(metaData_.declare_field<GenericFieldType>
+          =  &(metaData_.declare_field<double>
               (sideRank, "wall_friction_velocity_bip"));
-        stk::mesh::put_field(*wallFrictionVelocityBip, *part, numIp);
+        stk::mesh::put_field_on_mesh(*wallFrictionVelocityBip, *part, numIp, nullptr);
       }
     }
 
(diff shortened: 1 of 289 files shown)


spdomin:
Thank you for this PR. I plan on reviewing this on Friday. Let me know if the string change is part of this or you need time to push down this path. Ideally, we can do this in a single commit and we have the time to do so.

spdomin (changes requested):
@djglaze - this looks good to me. Lot's of modifications only by virtue of many many field registrations and "gets".

I only have a few questions/comments:

1. I noticed that you added a vector type for edge_area_vector to IO. I wonder if IO even supports edge field output? I seem to recall Nalu-Wind needing this for restart (mDot), however, never knew if this happened. @alanw0 may know. I am only curious to make sure that this IO change does not cause any issue if asking for an IO data type that is not supported (perhaps some nasty runtime throw).
2. Did you run the regression test suite on gcc 8.4-ish? I will make sure that I run the suite of tests before I accept the push.
4. Do we have any int registrations? I think that nalu_global_id is a size_t... I probably need to find this registration, although you may recall this nuance:) Can STK take double, int, size_t/uint_64, etc? That reminds me about size_t and uint_64... We need to check this on a Mac platform.
5. I need to check (again) to make sure that you added the set_field_output_type for all of the vector additions. In theory, this is the only field type in Nalu that needs this, right? All scalars and generics are fine. Did you do this by hand, or is there a more automated way to make sure VectorFieldTypes now have this "vector_3d" call?
6. As noted in 1 above, this all looks like standard, expected changes. As such, I appreciate your long overview of the motivation. I was simply wondering if you find that there is still value in the VectorFieldType, ScalarFieldType, and GenericFieldType abstraction? In the end, the are the same "types", i.e., double... As such, these types may only help readability and will not provide added compiler-time safety. In other words, you can always index a ScalarFieldType in the wrong manner. No need to change anything, just asking for your opinion here.. I am fine with the three types - even if for visual aesthetics:)

Let me know when Trilinos/develop is ready for this change and I can manage it. As such, the only "requested change" is to tell me when this push is ready for primetime. This way, everyone knows to avoid the push so that we do not break the nightly process.

djglaze:
> 1. I noticed that you added a vector type for edge_area_vector to IO. I wonder if IO even supports edge field output? I seem to recall Nalu-Wind needing this for restart (mDot), however, never knew if this happened. @alanw0 may know. I am only curious to make sure that this IO change does not cause any issue if asking for an IO data type that is not supported (perhaps some nasty runtime throw).

I don't exactly know about edge `Field` output and whether that works.  The call to `stk::io::set_field_output_type()` only adds an attribute that hangs off the side of the `Field` itself.  This is queried by IO only if you try to write the `Field` out.  The behavior under the hood is identical to what IO would have done with this `Field` before this change, where it would have sniffed the template parameter to get the "shape" of the `Field`.  If it worked before, it will continue to work now.  If it didn't work before, then it still won't work.

> 2. Did you run the regression test suite on gcc 8.4-ish? I will make sure that I run the suite of tests before I accept the push.

I was only able to piece together the modules and TPLs to get Nalu to build on gcc-7.2 with a modest amount of effort.  Under gcc-7.2, I have confirmed that all unit tests and all regression tests are clean.  I have not formally checked Nalu under gcc-8.4, but these changes are simple enough that I'm not expecting trouble.  Once we get these changes into Sierra (which might be today, now that our CI testing issues are resolved), the nightly testing will run these changes against Sierra apps on gcc-8.4.  If that's clean, then I'm confident that Nalu will be clean, also.

> 3. Do we have any int registrations? I think that nalu_global_id is a size_t... I probably need to find this registration, although you may recall this nuance:) Can STK take double, int, size_t/uint_64, etc? That reminds me about size_t and uint_64... We need to check this on a Mac platform.

Yes, there were a few `uint64_t` types in there related to your `GlobalIdFieldType`.  STK `Field`s are formally tested against types of: `char`, `unsigned char`, `signed char`, `short`, `unsigned short`, `int`, `unsigned int`,
`long`, `unsigned long`, `long long`, `unsigned long long`, `float`, `double`, `long double`,
`std::complex<float>`, and `std::complex<double>`.  The `uint64_t` type should map to one of these on the mac, so it should be just fine.

> 4. I need to check (again) to make sure that you added the set_field_output_type for all of the vector additions. In theory, this is the only field type in Nalu that needs this, right? All scalars and generics are fine. Did you do this by hand, or is there a more automated way to make sure VectorFieldTypes now have this "vector_3d" call?

Yes, Nalu only needed the extra type mapping call on vectors.  I've triple-checked it all at this point, so I'm pretty confident that it's correct.  Another set of eyes couldn't hurt, though.  :-)  And, it was indeed all done by hand.  It took about 2 weeks of work in my spare time.  It probably could have been automated some, but I wanted to visit each call site of interest for these changes and make sure that all substitutions were made correctly in the surrounding code context and make sure that the code formatting was consistent with what was already there.

> 5. As noted in 1 above, this all looks like standard, expected changes. As such, I appreciate your long overview of the motivation. I was simply wondering if you find that there is still value in the VectorFieldType, ScalarFieldType, and GenericFieldType abstraction? In the end, the are the same "types", i.e., double... As such, these types may only help readability and will not provide added compiler-time safety. In other words, you can always index a ScalarFieldType in the wrong manner. No need to change anything, just asking for your opinion here.. I am fine with the three types - even if for visual aesthetics:)

That's an interesting question.  You're right that there is now no difference between any of the double-flavored `Field` types.  We could certainly drop these duplicate typedefs.  The access patterns for a scalar, vector, or generic array are, and always have been, decoupled from the `Field` type as you know.  Keeping the "shape" embedded in the name for the `Field` type does still provide a visual reminder for how the `Field` should be accessed.  You're right that it's all aesthetics.  I think I prefer keeping them.

It's still going to be a while before these changes are in Trilinos/develop.  Alan's been struggling with Trilinos' broken auto-tester for the last week that's been blocking a push of his.  I don't know the current status, but I'm guessing it's still broken and I don't think there's an ETA yet.  I'll let you know when we're past the log-jam.

Dave

spdomin:
Thank you for the detailed set of replies. Looking forward to committing this set of changes. Best of luck on the merge to develop.

spdomin:
I noticed one fail due to a clash in the recent ExplicitFiltering class implementation. I had a list of clean-up anyway, and will accept this PR and fix this issue in my separate cleanUp project.

Error: Invalid call to MetaData::declare_field() for Field 'explicit_filter'.
  Cannot use a Field as the template parameter:
    declare_field<stk::mesh::Field<double, void, void, void, void, void, void, void>>()
  Please use only a datatype template parameter, e.g. declare_field<double>().
  Called from: /home/spdomin/gitHubWork/NaluDGlaze/src/ExplicitFiltering.C:177

PR #594, opened 2022-10-27
Support new thermal source term (and dep cleanup)

spdomin:
* add external data flag for walls; allows for one-way thermal

* deal with a couple Kokkos deprecation warnings

* deal with STK deprecation warnings


spdomin:
No new field registration. As such, no worry about merge conflict with STK field PR that is waiting on a clean STK development version.

PR #596, opened 2022-11-01
Explicit filtering

spdomin:


spdomin:
Used new-style of field registration and management. As such, there should be no conflict with @dglaze

PR #653, opened 2023-07-12
GCL and deform_wrt_model_coordinates fix

spdomin:
* missing scalarNp1_ assignment in scalar GCL.

* Cleaned up the elastic mesh motion kernel to include deform_wrt_model_coordinates. This was missing in the conversion to consolidated.

* plumbing for deform_wrt_model_coordinates in EqSys; activated in movingCylinder

* Remove activate_mass in MeshDisplaceEQS  since this is managed by the element source term activation.

* rebless due to deform_wrt_model_coordinates and Z-transport with GCL.

Notes:

a) physical density can be useful to specify when all of the mesh
   volumes around a moving boundary are the same, e.g., the
   movingCylinder case runs smoother  when rho = 0.005. However,
   convergence is good with the current geometric scaling option.


spdomin:
@wjhorne, 'should be the last FYI on the conversion effort for mesh_motion kernels.

PR #664, opened 2023-09-25
Modify and update CMake

spdomin:
* as per https://github.com/trilinos/Trilinos/issues/12326

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 ############################ BASE ######################################
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 3.23.0)
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 cmake_policy(SET CMP0057 NEW)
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -13,6 +13,11 @@ option(ENABLE_PARAVIEW_CATALYST
 option(ENABLE_WARNINGS "Add -Wall to show compiler warnings" ON)
 option(ENABLE_EXTRA_WARNINGS "Add -Wextra to show even more compiler warnings" OFF)
 
+########################### NALU-first #####################################
+# Set Nalu's compilers, CMAKE_FIND_LIBRARY_PREFIXES
+# and CMAKE_FIND_LIBRARY_SUFFIXES, etc, after inheriting settings from Trilinos
```
bartlettroscoe, line 18: "after inheriting settings from Trilinos" is not really true anymore, correct?

spdomin: You mean that second line?

bartlettroscoe: What I means is that NaLU is no longer "inheriting settings from Trilinos" since it has to define the compilers **before** calling `find_package(Trilinos ...)`.

PR #672, opened 2023-11-17
Clean-up compiler warnings, six-DOF, reset

spdomin:
* fix one-API warnings

* add reset option (tested; diff accepted)

* Missing post-processing of yplus/tauwall/pressure_force for six-DOF (found via above warning effort; added nodal field registraiton and population).

* define a new flag, meshMotionIncludesSixDof_ to ensure we do not process algorithms when six-DOF is not active (while general mesh motion is).

* reset time to allow a user to have a clean new time for precursor use cases.


spdomin:
@wjhorne - just a FYI on six-DOF post-processing changes and processing these algs only when six-DOF was activated (not just when mesh motion is active).

PR #674, opened 2024-01-25
Parallel and peridoc consistency in IC algs

spdomin:
* For periodic flows, you need perfect consistency between periodic node pairs.

Diffs accepted in periodic cases (very small)


spdomin:
@wjhorne and @psakievich : This fix may or may not be germane to Nalu-Wind. However, we had a few channel runs (WMLES) where a slight inconsistency in the IC specified between periodic nodes caused noise to persist. Best

psakievich:
Thanks @spdomin. This is good to be aware of. I appreciate you reaching out to let us know.

PR #694, opened 2024-09-27
new golds due to STK remove sort change

spdomin:
stkSortDiffCause


spdomin:
- HC diffs due to GL quadrature and multiple valid elements served up; fluids case also had equally valid coarse search elements due to the unique intersection of two Tri3 faces across a Hex8 "current" exposed surface.
- No need to exercise new interface to force a sort on the STK side.

spdomin:
All done, @alanw0.

